PR #46, opened 2019-03-14
Update README with activatability & rewordings

rakina:
cc @fergald @vmpstr @chrishtr @domenic

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,8 +88,12 @@ async function updateDom() {
   // Calling update() will allow rendering to go update the style & layout
   // values through co-operative updates that may last multiple frames.
   // Note that this will not cause the element to be painted/rendered.
-  // After the calculations finish the promise it returns will resolve.
+  // Also, only elements in the locked element's subtree will be updated
+  // co-operatively. If there are other things that needed rendering update
+  // in the document, those things will be updated in one go.
   let updatePromise = element.displayLock.update();
+
+  // After the calculations finish the promise update() returns will resolve.
   updatePromise.then(() => {
     // Calling commit() will cause the element to be rendered on the next frame.
     // Returns a promise that resolves when rendering finishes.
--- a/README.md
+++ b/README.md
@@ -23,8 +23,8 @@ DOM at once in order to prevent jank. As an example, ReactJS is [adding the
 ability](https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html)
 to do async operations to help with this problem.
 
-Creating more complex applications might also mean having a **bigger amount of things in a web page**,
-making the rendering costs to have all of the content to actually be part of the DOM prohibitively expensive.
+Creating more complex applications might also mean having a **larger amount of things in a web page**,
+making the rendering costs of having all of the content to actually be part of the DOM prohibitively expensive.
 
 This drives some amount of web authors to *virtualization* instead,
 where they actually **refrain from putting things in the DOM**,
@@ -32,9 +32,9 @@ keeping content in memory in some non-DOM data structure,
 with only the visible portion converted into DOM nodes and inserted into the document,
 recycling them in and out as needed.
 
-However, this will **make the web app suffer** because a lot of things like find-in-page,
+However, this will **cause problems to the web app** because a lot of things like find-in-page,
 accessibility, indexability, focus navigation, anchor links, etc.
-**depends on having things in the DOM**.
+**depends on having the content be in the DOM**.
 
 ### Proposal
 
@@ -46,20 +46,21 @@ If an element is display locked, it means that
 **any DOM, style or layout updates to its subtree are not rendered immediately**.
 Instead, when allowed, the user agent will process rendering updates *co-operatively*,
 yielding periodically to allow script or user interactions to happen.
-The web author can also commit/unlock the lock,
+The web author can also commit the lock,
 which triggers a synchronous update and render when necessary,
 causing the element to get updated and rendered in the next frame.
 
-The visual content of an element's subtree which is locked for display does not
-change. When locked, if the element already existed in the DOM, then the
-content that was present at the time the lock was acquired remains.
+When locked, if the element already existed in the DOM,
+then the visual content that was present at the time the lock was acquired is cleared,
+leaving only the space it takes in layout,
+similar to an element with `visibility: hidden;`.
 
 If the element was locked before being inserted into the DOM, then it is
 effectively inserted in a hidden state, which means the user agent does not
 display any content for this element or its subtree until the lock gets committed.
 
-The web author can also opt-in to allow display-locked elements to be **committed by the user agent in cases like focus navigation,
-find-in-page match navigation**, etc.
+The web author can also opt-in to allow display-locked elements to be
+**committed by the user agent in cases like focus navigation, find-in-page match navigation**, etc.
 so that they are properly rendered when needed in those cases.
 
 ### Example
@@ -72,8 +73,8 @@ async function updateDom() {
   element.appendChild(...);
   element.id = "...";
 
-  // Acquire a lock, and allow it to be unlocked by the user agent when
-  // necessary.
+  // Acquire a lock, and allow it to be committed by the user agent when
+  // necessary when focused, navigated to in find-in-page, anchor links, etc.
   await element.displayLock.acquire({ activatable: true });
 
   // Append the element to the DOM.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,21 +26,28 @@ to do async operations to help with this problem.
...
-locking an element.
+slow rendering phase updates.
+Specifically, we propose a concept of *display locking* an element.
 
-If an element is locked for display, it means that any DOM updates to its
-subtree are not rendered immediately. Instead, when the lock is committed, the
-user-agent processed rendering updates co-operatively, yielding periodically to
-allow script or user interactions to happen.
+If an element is display locked, it means that
+**any DOM, style, layout updates to its subtree are not rendered immediately**.
```
fergald:
```suggestion
**any DOM, style or layout updates to its subtree are not rendered immediately**.
```

README.md
```diff
@@ -52,38 +59,51 @@ async function updateDom() {
...
 
-  // Now we can update the element, causing co-operative updates. After that
-  // resolves, we can commit the element making it visible to the user.
-  element.displayLock.update().then(() => {
+  // Now even if the element is appended to the tree we can do DOM updates,
+  // change style, etc. to it but it won't affect rendering.
+  element.style = "...";
+  element.appendChild(...);
+
+  // Calling update() will allow rendering to go update the style & layout
```
fergald: I think the correct thing to say here is that style and layout inside the locked element will be updated (cooperatively) but style and layout outside the locked element will not.

README.md
```diff
@@ -52,38 +59,51 @@ async function updateDom() {
...
+  // values through co-operative updates that may last multiple frames.
+  // Note that this will not cause the element to be painted/rendered.
+  // After the calculations finish the promise it returns will resolve.
+  let updatePromise = element.displayLock.update();
+  updatePromise.then(() => {
+    // Calling commit() will cause the element to be rendered on the next frame.
```
fergald: This isn't quite accurate. In a flex situation, there could be a lot of work to be done after commit that we could not do during update and it could take more than a frame to do it.

I think we should document clearly somewhere (maybe not here, maybe just reference it from here) that update does its best to calculate everything and make commit cheap but there are situations like flex where this is impossible.

README.md
rakina: cc @vmpstr to make sure

README.md
rakina: cc @vmpstr to make sure

domenic:
It might be good to explicitly list the activation scenarios?

rakina:
> It might be good to explicitly list the activation scenarios? Otherwise looks good from my perspective.

Do you mean a list of things that trigger activation, like the one on the [explainer](https://github.com/WICG/display-locking/pull/47/files#diff-817a23c792520ce222813d3eff70d031R425)?

domenic:
Ah yes, I didn't realize there were separate explainer and README documents (and corresponding PRs). My bad!

README.md
```diff
@@ -23,24 +23,44 @@ DOM at once in order to prevent jank. As an example, ReactJS is [adding the
...
+Instead, when allowed, the user agent will process rendering updates *co-operatively*,
+yielding periodically to allow script or user interactions to happen.
+The web author can also commit/unlock the lock,
+which triggers a synchronous update and render when necessary,
+causing the element to get updated and rendered in the next frame.
 
 The visual content of an element's subtree which is locked for display does not
-change. Specifically, if the element already existed in the DOM, then the
-content that was present at the time the lock was acquired remains. 
+change. When locked, if the element already existed in the DOM, then the
```
rakina: So AFAIK "content that was present at the time the lock was acquired remains" is not true, at least in the current impl of Display Locking. I think a while ago it was determined that V1 of Display Locking won't retain the painted output for lock-after-append case. Is it still the case, and if so should we change this line too? @chrishtr @vmpstr

README.md
chrishtr: Fergal's comment is correct and should be noted.

README.md
chrishtr: Correct, we plan to skip this behavior in v1. Ok to remove.

rakina: Changed the line to say it's like `visibility: hidden`

README.md
rakina: Added a note that it might still take a long time in cases like flex

README.md
```diff
@@ -9,8 +9,8 @@ Websites frequently rely on dynamic content to present information. This means
 that DOM is often manipulated using script to present rich and dynamic content.
 There are cases when this can cause slow rendering phase updates, including
 script to update views, style updates, layout, and paint. This, in turn, causes
-jank or noticeable delay when presenting content, because rendering phase is
-updated synchronously with user interactions and requestAnimationFrame script.
+**jank or noticeable delay** when presenting content, because rendering phase is
```
chrishtr:
```suggestion
**jank or noticeable delay** when presenting content, because the rendering phase is
```

README.md
```diff
@@ -23,24 +23,45 @@ DOM at once in order to prevent jank. As an example, ReactJS is [adding the
 ability](https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html)
 to do async operations to help with this problem.
 
-### Proposal
+Creating more complex applications might also mean having a **larger amount of things in a web page**,
```
chrishtr:
```suggestion
Creating more complex applications might also mean having a **larger number of things on a web page**,
```

README.md
```diff
@@ -23,24 +23,45 @@ DOM at once in order to prevent jank. As an example, ReactJS is [adding the
...
+making the rendering costs of having all of the content to actually be part of the DOM prohibitively expensive.
 
-This document proposes to augment user-agent APIs to help address the problem of
-slow rendering phase updates. Specifically, we propose a concept of display
-locking an element.
+This drives some amount of web authors to *virtualization* instead,
```
chrishtr:
```suggestion
This drives some number of web authors to *virtualization* instead,
```

README.md
```diff
@@ -23,24 +23,45 @@ DOM at once in order to prevent jank. As an example, ReactJS is [adding the
...
+where they actually **refrain from putting things in the DOM**,
+keeping content in memory in some non-DOM data structure,
+with only the visible portion converted into DOM nodes and inserted into the document,
+recycling them in and out as needed.
 
-If an element is locked for display, it means that any DOM updates to its
-subtree are not rendered immediately. Instead, when the lock is committed, the
-user-agent processed rendering updates co-operatively, yielding periodically to
-allow script or user interactions to happen.
+However, this will **cause problems to the web app** because a lot of things like find-in-page,
```
chrishtr:
```suggestion
However, this will **cause problems to the web app** because a lot of user-agent features like find-in-page,
```

README.md
```diff
@@ -23,24 +23,45 @@ DOM at once in order to prevent jank. As an example, ReactJS is [adding the
...
+accessibility, indexability, focus navigation, anchor links, etc.
+**depends on having the content be in the DOM**.
```
chrishtr:
```suggestion
**depend on having the content be in the DOM**.
```

README.md
```diff
@@ -23,24 +23,45 @@ DOM at once in order to prevent jank. As an example, ReactJS is [adding the
...
 
-The visual content of an element's subtree which is locked for display does not
-change. Specifically, if the element already existed in the DOM, then the
-content that was present at the time the lock was acquired remains. 
+### Proposal
+
+This document proposes to augment user-agent APIs to help address the problem of
+**slow rendering phase updates**, and having some control on **when to pay rendering costs**
```
chrishtr:
```suggestion
**slow rendering phase updates**, and having some control on **when to pay rendering costs**.
```

README.md
chrishtr: In this case, which is lock-before-append, style, layout etc of the element and its subtree will be updated cooperatively, 
but the rest of the DOM will be updated in one go.

README.md
```diff
@@ -52,38 +73,55 @@ async function updateDom() {
...
+  let updatePromise = element.displayLock.update();
+  updatePromise.then(() => {
+    // Calling commit() will cause the element to be rendered on the next frame.
+    // Returns a promise that resolves when rendering finishes.
+    // Note that we can call commit() without calling update() beforehand, but
+    // that might cause a jank because we're doing the style, layout, paint in
+    // one frame instead because we don't have the updated values.
+    // However, even if we call update() beforehand, there are cases like flex
+    // layout where the layout of the locked element affects its siblings,
+    // possibly causing this to be janky.
```
chrishtr:
```suggestion
    // possibly causing this to be slow.
```

README.md
rakina: Added more comments.

README.md
```diff
@@ -52,38 +73,60 @@ async function updateDom() {
...
+  // change style, etc. to it but it won't affect rendering.
+  element.style = "...";
+  element.appendChild(...);
+
+  // Calling update() will allow rendering to go update the style & layout
+  // values through co-operative updates that may last multiple frames.
+  // Note that this will not cause the element to be painted/rendered.
+  // Also, only elements in the locked element's subtree will be updated
+  // co-operatively. If there are other things that needed rendering update
+  // in the document, those things will be updated in one go.
```
fergald:
```suggestion
  // in the document, those things will be updated in one go when the lock is committed.
```

rakina: Other dirty parts of the document that aren't in a locked subtree will be updated as part of the usual rendering updates, not when the lock gets committed.

I updated the comments around here again

rakina:
> Actually, can we remove large chunks of the README where they are just repeating the explainer?

Removed some parts of the problem & proposal, should we remove more?

PR #47, opened 2019-03-14
Update explainer with activatable & rewrite some parts

rakina:
cc @fergald @vmpstr @chrishtr @domenic

Not sure about the example that I added, also this only adds "activatable", will add contentSize in another PR probably.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -26,7 +26,7 @@ previous layout and visual update to be done before running subsequent script.
 Creating more complex applications might also mean having a **larger number of things in a web page**,
 making the rendering costs of having all of the content to actually be part of the DOM
 prohibitively expensive.
-This drives some amount of web authors to *virtualization* instead,
+This drives some number of web authors to *virtualization* instead,
 where they actually **refrain from putting things in the DOM**,
 keeping content in memory in some non-DOM data structure,
 with only the visible portion converted into DOM nodes and inserted into the document,
@@ -348,7 +348,7 @@ and `false` if not.
 
 #### The locked state
 
-#### What happens
+##### What happens
 
 When its lock is acquired, if the element already existed in the DOM,
 the visual content that was present at the time the lock was acquired is cleared,
@@ -356,13 +356,14 @@ leaving only the space it takes in layout,
 similar to an element with `visibility: hidden;`.
 
 If the element was locked before being inserted into the DOM,
-then it is effectively inserted in an state similar to `display: none;`,
+when it got inserted to the DOM, it will be in an state similar to `display: none;`.
+See [lock-after-append](#lock-after-append-locked-subtree) vs [lock-before-append](#lock-before-append-locked-element--subtree).
 
 ##### Modifications to the locked subtree
 
 Changes to the DOM, style, layout, etc of the *locked element itself*
-will be applied immediately if the element is locked after it is attached to the DOM,
-but will be ignored if it is locked before it is attached to the DOM.
+will be applied immediately if the element is locked after it is inserted to the DOM,
+but will be ignored if it is locked before it is inserted to the DOM.
 See [lock-after-append](#lock-after-append-locked-subtree) vs [lock-before-append](#lock-before-append-locked-element--subtree).
 
 Changes to the *descendants of a locked element* updates the DOM in such a way that script can inspect it immediately,
@@ -726,14 +727,12 @@ is acquired
 #### Lock-after-append. (locked subtree)
 
 When the element's lock is acquired while the element is already a part of the
-DOM, then the draw commands associated with that element's subtree are stashed
+DOM, visual content of elements within the locked subtree
+(not the locked element itself) is cleared
+and the draw commands associated with that element's subtree are stashed
 and used while the lock is acquired. Note that in this mode, changes to the
 element itself (e.g. border, size) are updated synchronously. In other words,
 the element itself is not locked for display, only its subtree.
-
-The visual content of the locked subtree is cleared,
-leaving only the space occupied by the locked element,
-similar to having `visibility: hidden` on the locked element.
 
 This mode is appropriate to use when the element's subtree needs to be updated
 without jank.
--- a/explainer.md
+++ b/explainer.md
@@ -728,8 +728,8 @@ is acquired
 When the element's lock is acquired while the element is already a part of the
 DOM, visual content of elements within the locked subtree
 (not the locked element itself) is cleared
-and the draw commands associated with that element's subtree are stashed
-and used while the lock is acquired.
+and changes to the subtree will not be reflected on screen until the lock is committed
+and the corresponding promise resolves.
 
 Note that in this mode, the locked element is rendered normally and changes to the
 element itself (e.g. border, size) are updated synchronously. In other words,
--- a/explainer.md
+++ b/explainer.md
@@ -32,9 +32,9 @@ keeping content in memory in some non-DOM data structure,
 with only the visible portion converted into DOM nodes and inserted into the document,
 recycling them in and out as needed.
 Popular examples include https://m.twitter.com/ and [react-window](https://github.com/bvaughn/react-window).
-However, this **causes problems for the web app** because a lot of things like find-in-page,
+However, this **causes problems for the web app** because a lot of user-agent features such as find-in-page,
 accessibility, indexability, focus navigation, anchor links, etc.
-**depends on having the contents in the DOM**.
+**depend on having the contents in the DOM**.
 
 We propose a new concept, **display locking**, to assist developers with **alleviating
 jank caused by DOM updates** and having some control on **when to pay rendering costs**,
@@ -158,7 +158,7 @@ updates which **induce large document lifecycle updates**, including style, layo
 compositing, and paint. In turn, these can cause jank on the page, due to
 lifecycle updates being synchronous with script and user interactions.
 
-In a large enough web app, even having small updates like window resizing or a style tweak might be expensive,
+In a large enough web app, even small updates like window resizing or a style tweak might be expensive,
 due to the **large amount of things getting affected**,
 even though we might **not necessarily need every part of the web app to be up-to-date all the time**. 
 
@@ -276,7 +276,7 @@ remainingItems.forEach(item => {
     let itemEl = createElementForItem(item);
     itemEl.displayLock.acquire({ timeout: Infinity, activatable: true });
     itemsList.appendChild(itemEl);
-    // We can do expensive operations to the items without worrying of the rendering costs.
+    // We can do expensive operations to the items without worrying about the rendering costs.
     // After we finished all the operations we can trigger a co-operative update & commit.
     // We might do this in a fancier way by not actually committing the element,
     // leaving it locked and detecting when to commit by using IntersectionObservers etc,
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -23,17 +23,33 @@ if the user agent is busy performing layout. The reason for this is that script
...
+
+We propose a new concept, **display locking**, to assist developers with **alleviating
+jank caused by DOM updates** and having some control on **when to pay rendering costs**,
+while also **allowing things that depend on the updates to get the up-to-date values** when urgent.
+
+Using display locking, the developer will be able to lock an element and its subtree,
+preventing visual updates.
+Then, the developer will be able construct the locked subtree’s DOM however they desire,
+and insert it into the DOM without any rendering cost or jank. After insertion, the element
+can be updated. These updates will be *co-operative* -- **interleaved with other
```
fergald: "updated" here needs some explanation. Maybe "marked for update".

fergald:
```suggestion
can be updated. These updates can be *co-operative* -- **interleaved with other
```

If they are small, they will complete immediately?

explainer.md
```diff
@@ -23,17 +23,33 @@ if the user agent is busy performing layout. The reason for this is that script
...
+work such as running script or DOM updates outside of the locked subtree**. The
 developer will then be able to commit the element's lock, causing the visual
 updates of the modified subtree to appear.  In essence, display locking will
-make it possible to perform complicated DOM updates without causing the rest of
-the page to jank.
+make it possible to **perform complicated DOM updates without causing the rest of
+the page to jank** and **only pay rendering costs on things that really needs it**.
```
fergald:
```suggestion
the page to jank** and **only pay rendering costs on things that really need it**.
```

explainer.md
```diff
@@ -236,30 +242,62 @@ user.
...
+    // After we finished all the operations we can trigger a co-operative update & commit.
+    // We might do this in a fancier way by not actually committing the element,
+    // leaving it locked and detecting when to commit by using IntersectionObservers etc,
+    // but that's out of the scope of this example :)
+    doExpensiveOperationsToEl(itemEl).then(element.updateAndCommit);
+  });
+});
+
+// If we need the updated style / layout values of the element, we can do it co-operatively.
+function getOffsetTopForLockedEl(element) {
```
fergald: I'm a little confused by this function. If the element is locked but none of its ancestors are, then I think this will cause layout to run on the whole doc. Locking the element on makes the layout values of things inside it cooperative, right?

explainer.md
```diff
@@ -335,13 +403,68 @@ not yet resolved. This causes work that is still needed to become synchronous,
...
+
+#### Element activation
+
+When an element is locked,
+there are some actions in the page that might require the element to get unlocked and be rendered to work properly.
+
+If the browser needs to *activate* a locked element
+or an element that has one or more locked ancestors,
+in order to make the activated element be visible,
+it will commit all locked elements in the ancestor chain of the element to be activated,
```
fergald: I'm not sure if it's correct but maybe

Will commit all locked elements in the inclusive ancestor chain which have the `activatable` flag set to `true`

But I'm not sure if the bubbling jumps over or stops at activatable=false locked elements.

explainer.md
```diff
@@ -335,13 +403,68 @@ not yet resolved. This causes work that is still needed to become synchronous,
...
+| bubbles  | true  |
+| composed | false |
+| target  | the previously-locked ancestor |
+| activatedElement  | The element that needs activation  |
+
+
+Consider this tree structure:
+```html
+<div id="first">
+ <div id="second"> <!-- locked -->
```
fergald: might be good to include 1 that has activatable=false.

explainer.md
chrishtr: Locking an element when it is already in the DOM tree only locks its subtree, not the element.

explainer.md
```diff
@@ -144,29 +147,32 @@ suffer from similar drawbacks.
...
   the contents.
 
-In general, large-scale updates to application state in a web app can cause
-updates which induce large document lifecycle updates, including style, layout,
+In general, **large-scale updates** to application state in a web app can cause
+updates which **induce large document lifecycle updates**, including style, layout,
 compositing, and paint. In turn, these can cause jank on the page, due to
 lifecycle updates being synchronous with script and user interactions.
 
+In a large-enough web app, even having small updates might be expensive even though
```
chrishtr: I find this sentence confusing. Are you trying to simply say that if the document is very large, even small diffs to it can be expensive?

explainer.md
```diff
@@ -275,29 +313,59 @@ Acquire performs the following steps:
...
   * This is an optional timeout value for the duration of the lock. When the
-    lock is acquired, and commit is not called within timeoutInMilliseconds
+    lock is acquired, and commit is not called within `timeoutInMilliseconds`
     milliseconds, the lock will be automatically committed. This is a safety
     valve for protecting pages from inadvertently leaving an element locked
     indefinitely due to uncaught exceptions or other bugs in the code. For a
     more sophisticated use cases, the value of Infinity will effectively disable
     the timeout.
+* (optional) `activatable`: `true|false`
+  * Signifies whether the user-agent can automatically unlock/commit the lock
```
chrishtr: remove "unlock"

explainer.md
```diff
@@ -275,29 +313,59 @@ Acquire performs the following steps:
...
-##### DisplayLockContext.update()
+#### What happens
+
+The element and its subtree's paint output (i.e. the output of the
+layout and paint phase) will not change while the lock is acquired. In the
+lock-before-append case, this means that the visual state is empty. What this
+allows is for the user-agent to not have to finish processing the locked
+element’s subtree when processing the layout and paint phase.
+
+In other words, when requested, the user-agent is free to process parts of the subtree,
```
chrishtr: This paragraph is I think hard to understand for non-experts. Is it really needed?

explainer.md
chrishtr: We haven't considered this case I think. I added an agenda item to discuss at the
weekly meeting.

explainer.md
```diff
@@ -335,13 +403,68 @@ not yet resolved. This causes work that is still needed to become synchronous,
...
+For all the elements that we commit as part of element activation,
+we will send a `beforeactivate` event to it just before committing.
+
+#### beforeactivate event
+
+| property  | value  |
+|---|---|
+| bubbles  | true  |
+| composed | false |
+| target  | the previously-locked ancestor |
```
chrishtr: Nearest previously-locked ancestor?

explainer.md
rakina: Yeah that makes sense. Added!

explainer.md
rakina: Since there can be multiple ancestor, not just nearest, I used "previously-locked ancestor" instead (and removed "the"). Maybe there's a better way to phrase this?

explainer.md
rakina: If any of its inclusive ancestor is locked & non-activatable, the element can't be activated, since even if we unlock everything else, it still won't be visible.

explainer.md
rakina: Changed it a bit, and also the commit part. PTAL

explainer.md
chrishtr: Looks good to me.

explainer.md
```diff
@@ -23,17 +23,39 @@ if the user agent is busy performing layout. The reason for this is that script
...
+Creating more complex applications might also mean having a **larger number of things in a web page**,
+making the rendering costs of having all of the content to actually be part of the DOM
+prohibitively expensive.
+This drives some amount of web authors to *virtualization* instead,
+where they actually **refrain from putting things in the DOM**,
+keeping content in memory in some non-DOM data structure,
+with only the visible portion converted into DOM nodes and inserted into the document,
+recycling them in and out as needed.
+Popular examples include https://m.twitter.com/ and [react-window](https://github.com/bvaughn/react-window).
+However, this **causes problems for the web app** because a lot of things like find-in-page,
```
chrishtr:
```suggestion
However, this **causes problems for the web app** because a lot of user-agent features such as find-in-page,
```

explainer.md
```diff
@@ -23,17 +23,39 @@ if the user agent is busy performing layout. The reason for this is that script
...
+accessibility, indexability, focus navigation, anchor links, etc.
+**depends on having the contents in the DOM**.
```
chrishtr:
```suggestion
**depend on having the contents in the DOM**.
```

explainer.md
```diff
@@ -144,29 +153,33 @@ suffer from similar drawbacks.
...
   the contents.
 
-In general, large-scale updates to application state in a web app can cause
-updates which induce large document lifecycle updates, including style, layout,
+In general, **large-scale updates** to application state in a web app can cause
+updates which **induce large document lifecycle updates**, including style, layout,
 compositing, and paint. In turn, these can cause jank on the page, due to
 lifecycle updates being synchronous with script and user interactions.
 
+In a large enough web app, even having small updates like window resizing or a style tweak might be expensive,
```
chrishtr:
```suggestion
In a large enough web app, even small updates like window resizing or a style tweak might be expensive,
```

explainer.md
```diff
@@ -236,30 +249,66 @@ user.
...
+
+<script>
+// Remaining items in the list are in |remainingItems|.
+// We want to add the remaining list items asynchronously.
+remainingItems.forEach(item => {
+  requestIdleCallback((deadline) => {
+    let itemEl = createElementForItem(item);
+    itemEl.displayLock.acquire({ timeout: Infinity, activatable: true });
+    itemsList.appendChild(itemEl);
+    // We can do expensive operations to the items without worrying of the rendering costs.
```
chrishtr:
```suggestion
    // We can do expensive operations to the items without worrying about the rendering costs.
```

explainer.md
```diff
@@ -275,29 +324,59 @@ Acquire performs the following steps:
...
+This is a boolean value that will be `true`
+if the `Element` associated with this `DisplayLockContext` is in the locked state,
+and `false` if not.
+
+#### The locked state
+
+#### What happens
+
+When its lock is acquired, if the element already existed in the DOM,
+the visual content that was present at the time the lock was acquired is cleared,
```
chrishtr: This text about visibility: hidden is ok for now, but just noting that the discussion
we had on Monday about intrinsic sizing has bearing on it.

However, there is a problem in your description: the root element being locked in this
case is still painted and laid out. It's the subtree that is not painted.

rakina: Hmm yeah, actually removed the `visibility:hidden` part now, we should update the details about intrinsic sizing later.

explainer.md
```diff
@@ -560,13 +727,15 @@ is acquired
 #### Lock-after-append. (locked subtree)
 
 When the element's lock is acquired while the element is already a part of the
-DOM, then the draw commands associated with that element's subtree are stashed
+DOM, visual content of elements within the locked subtree
+(not the locked element itself) is cleared
+and the draw commands associated with that element's subtree are stashed
 and used while the lock is acquired. Note that in this mode, changes to the
```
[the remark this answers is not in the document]
chrishtr: The stashing referred to stale pixels. Can be removed.

rakina:
@chrishtr @vmpstr can either of you give write access to me? Just realized I can't merge without that

PR #53, opened 2019-04-16
Add size option & make explainer not focus on implementation-specific stuff

rakina:
* Adds size option & brief explanation
* Moves impl specific stuff, lifecycle explanation to different files
* Simplifies example
* Remove remaining references to stale pixels stuff
* Minor tidying

Maybe add these here, or maybe future PR:
* Possible difference in layout values while locked, maybe even specifically mention getBoundingClientRect, offsetTop etc?
* Mention possible bad performance with flex etc

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -1,6 +1,6 @@
 # Display Locking
 
-### Introduction
+## Introduction
 
 As the quality and performance of the user-agents progresses, developers are
 looking towards the web as a platform to deliver rich, visually appealing and
@@ -36,28 +36,9 @@ However, this **causes problems for the web app** because a lot of user-agent fe
 accessibility, indexability, focus navigation, anchor links, etc.
 **depend on having the contents in the DOM**.
 
-### Proposal: Display Locking
-
 We propose a new concept, **display locking**, to assist developers with **alleviating
 jank caused by DOM updates** and having some control on **when to pay rendering costs**,
 while also **allowing things that depend on the updates to get the up-to-date values** when urgent.
-
-Using display locking, the developer will be able to lock an element and its subtree,
-preventing visual updates.
-Then, the developer will be able construct the locked subtree’s DOM however they desire,
-and insert it into the DOM without any rendering cost or jank.
-
-After insertion, we can request the rendering values (style, layout, etc) to be updated.
-These updates can be *co-operative* -- **interleaved with other
-work such as running script or DOM updates outside of the locked subtree** if they are going to take a long time,
-or calculated synchronusly if we need it.
-The developer is also able to commit the element's lock,
-which will calculate the rendering values if it's not up-to-date,
-and cause the visual updates of the modified subtree to appear.
-
-In essence, display locking will
-make it possible to **perform complicated DOM updates without causing the rest of
-the page to jank** and **only pay rendering costs on things that really need it**.
 
 In the rest of the document, we describe the display locking concept in detail.
 We will first examine some motivating examples which we commonly observe in rich
@@ -76,8 +57,7 @@ and presentation to the screen as well as areas that display locking is geared
 to improve. 
 A brief explanation of possible display locking implementations is also [available](./implementation.md).
 
----
-### Motivating Examples
+## Motivating Examples
 
 Consider the following example.
 
@@ -142,7 +122,7 @@ the subtree** (it’s an iframe in this example), it still janks because **DOM
 updates are atomic.**
 
 
-#### Common patterns
+### Common patterns
 
 There are other common patterns that exhibit similar behavior, and as a result
 suffer from similar drawbacks.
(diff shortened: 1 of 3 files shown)


explainer.md
```diff
@@ -40,39 +40,24 @@ We propose a new concept, **display locking**, to assist developers with **allev
...
+We will then move on to describe the API changes required to
 implement display locking, with a description of each new functionality. With
 that in place, we will revisit the motivating examples to see how display
 locking can be applied in order to improve their performance. Finally, we will
-discuss possible display locking implementations, along with possible corner
-cases and limitations.
+discuss possible corner cases and limitations.
+
+In order to understand the display locking proposal details, a brief explanation
+of common phases in user-agent lifecycles is available [here](./lifecycle.md).
```
vmpstr: Will "./lifecycle.md" pick up the correct thing when this is merged?

explainer.md
```diff
@@ -165,98 +150,29 @@ even though we might **not necessarily need every part of the web app to be up-t
...
 
+The developer will be able to lock an element and its subtree,
+preventing visual updates.
+Then, the developer will be able construct the locked subtree’s DOM however they desire,
+and insert it into the DOM without any rendering cost or jank.
+
+After insertion, we can request the rendering values (style, layout, etc) to be updated.
+These updates can be *co-operative* -- **interleaved with other
+work such as running script or DOM updates outside of the locked subtree** if they are going to take a long time,
+or calculated synchronusly if we need it.
```
vmpstr: s/if we need it/if needed/

explainer.md
```diff
@@ -165,98 +150,29 @@ even though we might **not necessarily need every part of the web app to be up-t
...
+The developer is also able to commit the element's lock,
+which will calculate the rendering values if it's not up-to-date,
```
vmpstr: "if they are not up-to-date"

explainer.md
```diff
@@ -165,98 +150,29 @@ even though we might **not necessarily need every part of the web app to be up-t
...
+and cause the visual updates of the modified subtree to appear.
+
+In essence, display locking will
+make it possible to **perform complicated DOM updates without causing the rest of
+the page to jank** and **only pay rendering costs on things that really need it**.
```
vmpstr: ... only pay rendering cost if and when the developer decides to do it

explainer.md
```diff
@@ -269,68 +185,94 @@ In this example, we want to add a lot of items to a list in a non-janky way.
...
+  itemEl.innerText = item;
+  itemEl.appendChild(...);
+  itemEl.firstChild.style = "...";
+  doExpensiveOperationsToChildren(itemEl).then(element.updateAndCommit);
 });
 
 // Getting style/layout values within a locked subtree :
-// - If we access child.offsetTop directly, it will cause a forced synchronous update
+// - If we call getComputedStyle, etc. directly, it will cause a forced synchronous style update
 //   (see [what-forces-layout](https://gist.github.com/paulirish/5d52fb081b3570c81e3a)).
```
vmpstr: This part references layout, but the above talks about style. I think we should mention both

explainer.md
```diff
@@ -269,68 +185,94 @@ In this example, we want to add a lot of items to a list in a non-janky way.
...
 // - We can call update() on the locked element to trigger a co-operative update and get
 //   the value after the calculations finished instead.
-function getOffsetTopForLockedElementChild(child, lockedElement) {
-  return new Promise((resolve) => {
-    lockedElement.displayLock.update().then(() => { resolve(child.offsetTop) };
-  });
-}
+lockedElement.displayLock.update().then(() => {
+  // They're free!
```
vmpstr: Maybe: "These accesses are cheap, since the subtree is updated" or something. I want to convey somehow that "free" is only the user-agents best attempt at making it free

explainer.md
```diff
@@ -269,68 +185,94 @@ In this example, we want to add a lot of items to a list in a non-janky way.
...
+### The locked state
+
+#### The locked element
+
+When its lock is acquired, the locked element itself will still be rendered and laid out normally.
+Changes to the DOM, style, layout, etc of the *locked element itself* will be applied as it would on any other normal element.
+
+##### Sizing
+
+When locked, the locked element is treated similar to as if it has [size containment](https://www.w3.org/TR/css-contain-1/#containment-size) with some differences.
```
vmpstr: s/with some differences/. Specifically, the following demonstrates how the element is laid out:/

explainer.md
```diff
@@ -269,68 +185,94 @@ In this example, we want to add a lot of items to a list in a non-janky way.
...
+
+* If some style rule specifies the size of the locked element, we will use that size for the locked element.
+* If there are no style rule specifying the size of the locked element, we will use the size given in the `acquire` call's options.
+Instead of being treated as having no contents, the locked element will behave as if there is one child with the given *locked content size* as its intrinsic width and height, for sizing purposes.
+* If there are no size specified in the `acquire` call, the locked element will not take any space, similar to `display: none` elements instead.
```
vmpstr: I would say here that if no size is specified, then the element will behave as if it has "contain: size" specified. In other words, it would lay out as if it is empty.

This is in contrast to display:none which can't be sized by style. We do allow sizing by style. In the future, I think we're thinking of adding things like size: "auto" and size: "none", with the latter truly behaving as display none.

explainer.md
```diff
@@ -269,68 +185,94 @@ In this example, we want to add a lot of items to a list in a non-janky way.
...
+
+#### The locked subtree
 
-### Proposed APIs Details
+When its lock is acquired, the **visual content of the nodes of flat-tree descendants of the element is cleared**, similar to `display: none`.
 
-#### Element.displayLock
+Changes to the *flat-tree descendants of a locked element* updates the DOM in such a way that script can inspect it immediately,
```
vmpstr: Is flat-tree terminology common across user-agents or is it blink specific?

explainer.md
```diff
@@ -339,44 +281,17 @@ pairs:
   * Not setting this, or setting this to `false`,
     means this element and its descendants will be ignored in cases like
     find-in-page, anchor link navigation, focus navigation, etc.
+* (optional): `size`:  `[width, height]`
+   * Indicates the *locked content size* of the locked element. See [sizing](#sizing) section for more detail.
+   * If not specified, the locked element will not take any space, similar to `display: none` elements.
```
vmpstr: if not specified, it will be treated as if [0, 0] is specified

explainer.md
rakina: Yep! Tested on the add-size branch.

implementation.md
```diff
@@ -0,0 +1,109 @@
+# Implementation description
+
+Note to the reader: this describes the sample prototype implementation.
+The design choices here are not meant to represent final design decisions, but
+act as a proof of concept implementation in Blink
```
chrishtr:
```suggestion
act as a proof of concept implementation in Blink.
```

explainer.md
vmpstr: Sounds good

explainer.md
vmpstr: Yeah that's a fair point, keeping it as is is fine too maybe then "with some differences, outlined below" or something like that so the reader doesn't have to wonder what differences we're talking about here

explainer.md
vmpstr: Yep that sounds good

vmpstr:
lgtm

rakina:
Thank you for the reviews!

Issue #74, opened 2019-09-03
Enumeration of activation algorithms.


vmpstr:
We need a document to list out all of our activation algorithms. Also, it would be nice to explicitly call out scrolling as not being one of them.

rakina:
FYI here's the list:
* find-in-page active match
* `scrollIntoView()`
* `focus()` and other focus navigation
* anchor link navigation
* selections
* accessibility focus, press, etc

Issue #72, opened 2019-09-03
Deprecate explainer.md in favor of explainer-new.md


vmpstr:
Since we switched from the explainer, we need to update the repo landing page.

Also, we might want to update the detailed explainer since it can become confusing to have both rendersubtree and displayLock references.

rakina:
Is this assigned to @chrishtr?

vmpstr:
Assigned to whoever wants to grab it :)

PR #75, opened 2019-09-05
Use explainer-new.md as README.md, and add section on activation

rakina:
Fixes https://github.com/WICG/display-locking/issues/74 and https://github.com/WICG/display-locking/issues/72

Probably we should remove explainer.md too. Do we still want a lengthy explanation of APIs and motivations and detailed rendering explanations?


README.md
```diff
@@ -1,133 +1,111 @@
...
+
+## Example usage
+
+```html
+<div id=target rendersubtree="invisible" style="content-size: 200px 200px">...content...</div>
+<script>
+target.setAttribute('rendersubtree', ''); // makes #target render
+</script>
+```
+This div's subtree is not rendered (but the div itself is; this allows the div to show fallback or "loading..." affordances), and there is no need for the browser to do any rendering lifecycle phases for the subtree of the div. Custom element upgrades are not performed and resources are not loaded. The div lays out as if it had a single 200px by 200px child, which serves as a placeholder in order to take up the approximate layout size of the div's subtree. This allows page layout to be approximately correct, and preserves layout overflow size for scrolling. The browser may *not* render the content, even via a user-agent feature.
```
vmpstr: > Custom element upgrades are not performed and resources are not loaded.

I think we need to update this to mention other values of rendersubtree, such as holdupgrades, and holdresourceload (not sure those are the names we agreed on though). If this is a straight up rename, then feel free to punt

README.md
```diff
@@ -1,133 +1,111 @@
...
+```
+The div's subtree does not render to the screen, but when updateRendering is called, the browser does work in the background to prepare to render it quickly in the future. This includes loading external resources referred to in the subtree and custom element upgrades. It also may include running the rendering lifecycle steps for the subtree up to and including style, layout, paint and raster. When the returned promise resolves, reading layout or style-inducing properties on the subtree is expected to be fast. Changing the `rendersubtree` value to `visible` should also be fast.
+
+## Element activation by the user agent
+
+When an element is not rendered because it's a part of a not-rendered subtree caused by `rendersubtree`, there are some actions in the page that might require the element (and its ancestors) to be rendered to work properly. If all of the element's rendering ancestor whose `rendersubtree` attribute contains `invisible` also contains `activatable`, then the user agent can *activate* the element. This will change of the non-null rendersubtree value of all of its rendering ancestors to the empty string - causing the element and its ancestors to get rendered.
```
chrishtr:
```suggestion
When an element is not rendered because it's a part of a not-rendered subtree caused by `rendersubtree`, there are some actions in the page that might require the element (and its ancestors) to be rendered to work properly. If all of the element's rendering ancestors' `rendersubtree` attributes contain contain `activatable`, then the user agent can *activate* the element. This will change of the non-null rendersubtree value of all of its rendering ancestors to the empty string - causing the element and its ancestors to get rendered.
```

chrishtr:
Thanks for doing this!

README.md
rakina: Moved them to a new example now

rakina:
> > Probably we should remove explainer.md too. Do we still want a lengthy explanation of APIs and motivations and detailed rendering explanations?
> 
> I kind of like the detailed explainer, but it's a bit of a pain to go through and update for rendersubtree, I think. For now, maybe we can rename it to explainer-deprecated.md or something?
> 
> Approved for merge if you want to do separate PRs for that, or let me know if you make them in this pull.

Yeah it's nice to have something like that, probably  we should update it later. Renamed in this PR for now.

Issue #73, opened 2019-09-03
CSS content-size standalone explainer


vmpstr:
We need a standalone explainer for content-size. Since it's pretty related to display-locking, I feel like it should be in this repo.

PR #76, opened 2019-09-07
Add standalone explainer for content-size.

chrishtr:
#73

### DIFF
--- a/explainer-content-size.md
+++ b/explainer-content-size.md
@@ -1,0 +1,34 @@
+# The `content-size` CSS property
+
+This is an explainer for why the `content-size` property is needed. A draft of the spec for this property is [here](http://tabatkins.github.io/specs/css-content-size/). The corresponding spec issue is [here](https://github.com/w3c/csswg-drafts/issues/4229).
+
+`content-size` represents the intrinsic layout sizing of a DOM subtree. Intrinsic sizing is an input to layout of ancestors, and is an important input to the layout constraint algorithms of the web. As one example, if an element has [`width: max-content`](https://drafts.csswg.org/css-sizing-3/#valdef-width-max-content) CSS on it, then it is sized to the [max-content inline size](https://drafts.csswg.org/css-sizing-3/#max-content-inline-size), which is the maximum intrinsic width of elements in the subtree. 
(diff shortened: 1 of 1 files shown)


chrishtr:
@tabatkins

explainer-content-size.md
```diff
@@ -0,0 +1,34 @@
...
+
+In addition, the height of an overflow-scrolling region depends on the [layout overflow size](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flow_Layout/Flow_Layout_and_Overflow) of its content, which includes intrinsic sizing. This allows a web developer to avoid spending time rendering scrollable content which is not visible to the user, but still have it take up space in the scroller.
+
+See also the [main explainer](https://github.com/WICG/display-locking/blob/master/README.md) for display-locking for more motivating examples along these lines.
+
+## Examples of `content-size`
+
+[Here](https://wicg.github.io/display-locking/sample-code/contain-size-block-flow-examples.html) is a page demonstrating some simple block-flow layout use-cases.
+
+[Here](https://wicg.github.io/display-locking/sample-code/contain-size-flexbox-examples.html) is a page demonstrating a basic flexbox use-case.
```
vmpstr: The last example in this link should say "with min-width/min-height", and also the style of it has a typo where it say "minheight" instead of "min-height"

explainer-content-size.md
```diff
@@ -0,0 +1,34 @@
...
+Currently, such intrinsic sizes are determined implicitly from other factors. For example, a div with `width: 200px` has a `200px` intrinsic width that is communicated upwards in layout algorithms in order to provide sizing constraints to ancestors. Another example is that an image with a `width=200px` attribute will have a `200px` intrinsic width; if it instead has a `src` attribute pointing at an image resource that has a `200px` image resolution width, the image will also have a `200px` intrinsic width once the image is loaded.
+
+This proposal exposes these intrinsic sizes explicitly. (In that sense, one motivation for the `content-size` property could be [explaining](https://extensiblewebmanifesto.org) these algorithms. One practical application could be its use in testing their behavior.
```
vmpstr: It would be nice to add a blurb about size containment. Something along the lines of trying not to break the web compat, in the sense that if we do use children for layout (ie no size containment) then we want to rely on that. However, if we can guarantee that children layout doesn't matter (ie size containment) then we can safely use content-size in its stead

explainer-content-size.md
chrishtr: Done.

explainer-content-size.md
chrishtr: Added.

PR #90, opened 2019-10-22
Trivial typo fix

bgirard:


bgirard:
I also felt like this sentence was ambiguous:
`This will change all of the non-null rendersubtree value of all of its ancestors to the empty string - causing the element and its ancestors to get rendered.`

Consider explicitly stating that it applies to the element itself as well as its ancestors.

tabatkins:
I'd happily accept another typo fix converting that to link to `[=inclusive ancestors=]` (a [term from DOM](https://dom.spec.whatwg.org/#concept-tree-inclusive-ancestor) that covers the concept explicitly).

PR #95, opened 2019-11-18
Update cheatsheet

vmpstr:
Please take a look. 

https://chromium-review.googlesource.com/c/chromium/src/+/1922302 adds the margin

### DIFF
--- a/cheatsheet.md
+++ b/cheatsheet.md
@@ -25,6 +25,11 @@ with possibly out-of-date explainers.
       tab-navigation to viewport activation, rename content-size to
       intrinsic-size</td>
 </tr>
+<tr>
+  <td>2019-11-18</td>
+  <td>Specify a 50% margin on viewport activation, update intrinsic-size status
+      (behind a flag), update activation status (default)</td>
+</tr>
 </table>
 
 ## Terminology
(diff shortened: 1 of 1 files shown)


cheatsheet.md
```diff
@@ -84,11 +89,17 @@ property is set to `""`. Note that `skip-activation` implies
...
   activated (disabled by `skip-activation`)
+
 ## Current status (Chromium)
 
-* **Activation**: As of [this patch](https://chromium-review.googlesource.com/c/chromium/src/+/1853854),
-  activation is the default behavior, which can be disabled with
-  `skip-activation` or `skip-viewport-activation`. A `beforeactivate` event will be fired before showing the new content on-screen.
+* **Activation**: Activation is the default behavior, which can be disabled with
+  `skip-activation` or `skip-viewport-activation`. A `beforeactivate` event will
+  be fired before showing the new content on-screen.
```
chrishtr: This event behavior is not true - we instead post a task.

vmpstr: I think we just fire it: https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/display_lock/display_lock_context.cc?l=384

There is a posttask to deliver the intersection observation, is this what you mean? Most activations do it right as we're activating. We are going to change this based on our previous discussions, but as it stands now I think the text is more or less correct. WDYT?

chrishtr: We shouldn't fire sychronously during C++ execution, or there will be re-entrancy issues...

PR #111, opened 2020-01-24
Cleaned up documentation

vmpstr:
I've updated most of the files with the exception of README.md, which I'll get to on Monday. PTAL at the rest of them.


cheatsheet.md
```diff
@@ -11,98 +11,103 @@ with possibly out-of-date explainers.
...
+    containment specified.
+* **contain-intrinsic-size**: this refers to a complementary CSS feature called
+  contain-intrinsic-size, which allows the developer to specify a placeholder size
+  when size containment is present, as it would be when the subtree is locked.
+
+## User-agent algorithms
+
+This is a list of algorithms that can bring an element into view or otherwise
+cause the user-agent to unlock the elements. Note that unless explicitly stated
+otherwise, the language below uses 'unlock' as the action since that is the
```
rakina: Can you use another term instead? It is a bit confusing :D

cheatsheet.md
```diff
@@ -11,98 +11,103 @@ with possibly out-of-date explainers.
...
+  algorithm will resume and only consider the active match as valid if the
+  subtree is now unlocked. Otherwise, it will proceed with looking for other
+  matches outside of the current locked root.
 
 ### Accessibility
-* Nodes in locked subtree (except those with `skip-activation`) are included in the AX tree but are marked as offscreen and don't have layout values, and thus are exposed to assistive technologies.
-* AX selections, focus, scrolling/navigation to nodes in locked subtrees will activate the locked ancestors.
-* Can be disabled with `skip-activation`.
-* There's still an active discussion on this, please see https://github.com/WICG/display-locking/issues/102 for more.
+* Locked subtrees are omitted from the accessibility tree. Note that there may
```
rakina: For `render-subtree: invisible` they are still in the accessibility tree, no?

vmpstr: Done.

cheatsheet.md
vmpstr: Do you have any suggestions? I kind of want to avoid the word 'activate' since it's been used to mean unlock in many of our discussions so it would be confusing. I think this should be the same word as this signal that we want to add. However, I also don't know of a good name for it.

cheatsheet.md
```diff
@@ -11,98 +11,104 @@ with possibly out-of-date explainers.
...
+  algorithm will resume and only consider the active match as valid if the
+  subtree is now unlocked. Otherwise, it will proceed with looking for other
+  matches outside of the current locked root.
 
 ### Accessibility
-* Nodes in locked subtree (except those with `skip-activation`) are included in the AX tree but are marked as offscreen and don't have layout values, and thus are exposed to assistive technologies.
-* AX selections, focus, scrolling/navigation to nodes in locked subtrees will activate the locked ancestors.
-* Can be disabled with `skip-activation`.
-* There's still an active discussion on this, please see https://github.com/WICG/display-locking/issues/102 for more.
+* Subtrees in the `render-subtree: invisible` configuration are included in the
```
chrishtr: Add the tentative design decision: locked content is not present in the AT. unlocked content is present.

cheatsheet.md
```diff
@@ -11,98 +11,104 @@ with possibly out-of-date explainers.
...
+  technology.  These algorithms should, for the most part, behave consistently
+  with non-accessibility ways of bringing content into view.
 
 ### Anchor link navigation:
-* When fragment link (ie url.html#elementid) navigation results in a navigation to an element in a locked subtree, its locked ancestors will be activated.
-* Can be disabled with `skip-activation`.
+* When fragment link (ie url.html#elementid or url.html#:~:text=foo) navigation
+  results in a navigation to an element in a locked subtree, its locked
+  ancestors will be unlocked. Tentative design: Similar situations as in
+  find-in-page cae should be considered. In particular, it may be prudent to
```
chrishtr:
```suggestion
  find-in-page case should be considered. In particular, it may be prudent to
```

chrishtr: Why say "prudent to" if it's a tentative decision anyway? Just say it should yield and then apply the scroll in a subsequent rendering task.

explainer-contain-intrinsic-size.md
```diff
@@ -0,0 +1,66 @@
...
+which is the maximum intrinsic width of elements in the subtree.  Currently, such
+intrinsic sizes are determined implicitly from other factors [2], and the other
+factors have additional undesirable layout side-effects.
+
+Under size containment, the intrinsic size is determined without consideration
+of element's children. This means that typically the intrinsic size will be 0x0,
+with some exceptions such as `display: grid` which also considers its tracks
+when determining the intrinsic size. 
+
+The contain-intrinsic-size extends size containment by allowing the developer to
```
chrishtr:
```suggestion
The `contain-intrinsic-size` property extends size containment by allowing the developer to
```

vmpstr:
PTAL.
- Addressed review feedback
- Finished updating README.md

cheatsheet.md
rakina: Maybe we can just keep "activate" but explain here? Or maybe, "unlock or send activation signal (depending on the render-subtree value)"?

vmpstr: I'm going to commit this as is since there's a bunch of other updates that should go in. We can follow-up in an issue about what to call this action.

README.md
```diff
@@ -94,151 +88,143 @@ Whether or not this is the final proposed set of features is yet undecided.
...
+      activation event (described below).
+
+  It is also worth noting that when the element is not rendered, then
+  `contain: layout style size;` is added to its style to ensure that the
+  subtree content does not affect elements outside of the subtree.
+  Furthermore, when the element is rendered in the `render-subtree: invisible`
+  configuration (i.e. the user-agent decides to render the element), then
+  `contain: layout style;` applies to the element.
+
+2. A new `contain-intrinsic-size` CSS property (spec draf to be added soon). This
```
cbiesinger: I'm late, but I noticed a typo (draf -> draft)

README.md
```diff
@@ -94,151 +88,143 @@ Whether or not this is the final proposed set of features is yet undecided.
...
-user-agent features in cases when it is not actually rendered to the user in the
-current application view.
+subtrees to render. Note that with sufficient containment and intersection
+observer, the functionality provided by `render-subtree` may be mimicked with
+some exceptions: find-in-page functionality does not work in unrendered content;
+this relies on more browser heuristics to ensure contained invisible content is
+cheap -- `render-subtree` is a stronger signal to the user-agent that work
+should be skipped.
+
+Similar to `visibility: hidden`, `contain: strict` allows the browser to
```
cbiesinger: I don't think vis:hidden allows that? Descendants of vis:hidden can be vis:visibile

README.md
chrishtr: This was fixed in a separate commit.

README.md
chrishtr: Yep, true. I'll add a qualification.

PR #116, opened 2020-02-18
Updated spec

vmpstr:
Please take a look. My plan is to have another version for contain-intrinsic-size. The yet-to-be-named event is likely to be a part of the html spec so I'll also add it in some other format.

Let me know if my changes here make sense and whether the plan is fine as well.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -91,14 +91,14 @@ Definitions {#definitions}
 ===========
 
 * <dfn export>off-screen</dfn>: an element is considered to be off-screen if
-    its border box does not intersect the viewport with a fixed, user-agent
+    its border box does not intersect the visual viewport plus a user-agent
     specified, margin. 
     <div class=note>
-      The the user-agent is free to consider a fixed margin
-      around the viewport when computing whether the element is off-screen. This
-      means that when an element is approaching the viewport, the user-agent can
-      start processing it in preparation of the element entering the
-      viewport. A 50% viewport margin is a reasonable implementation default
+      The the user-agent is free to consider any margin around the viewport
+      when computing whether the element is off-screen. The margin is meant to
+      allow the user-agent to start preparing the element ahead of time, before
+      it enters the viewport. A 50% viewport margin is a reasonable
+      implementation default
     </div>
 * <dfn export>on-screen</dfn>: an element is considered to be on-screen if it
     is not [=off-screen=]
@@ -148,15 +148,14 @@ Inherited: no
          etc.
 </dl>
 
-When element is [=skipped=], it is possible to avoid parts of rendering in the
-element's subtree. In these situations, the user-agent should avoid as much
-rendering work as possible. This also means that the user-agent should retain
-the previously computed layout state if possible.
+Note that when the element is [=skipped=] and the rendering work is avoided,
+the user-agent should retain the previously computed layout state if possible.
 
 <div class=note>
   Intuitively, a [=hidden=] value means that the element behaves in a way that
-  does not expose the contents of its subtree to the user. Script interactions
-  are required for the content to appear to the user.
+  does not expose the contents of its subtree to the user (but still has an
+  effect on layout, especially in conjunction with `contain-intrinsic-size`).
+  Script interactions are required for the content to appear to the user.
 
   In contrast, a value of [=auto=] means that the element behaves as if the
   contents of the subtree are accessible to the user. The content can be
@@ -219,16 +218,34 @@ Examples {#examples}
 <div class=example>
 	<xmp highlight=markup>
 		<style>
+    body {
+      margin: 0;
+    }
 		.sv {
 		  subtree-visibility: hidden;
+
+      position: relative;
+      left: 10px;
+      top: 20px;
 		}
+    #child {
+      position: relative;
+      left: 1px;
+      top: 2px;
+
+      width: 100px;
+      height: 200px;
+    }
 		</style>
 
 		<div id=target class=sv>
-		  ... some content goes here ...
+      <div id=child></div>
+		  ... some other content goes here ...
 		</div>
     <script>
       ...
+      // This will force rendering work, including layout,
+      // if the UA previously avoided it.
       target.firstElementChild.getBoundingClientRect();
       ...
     </script>
@@ -239,7 +256,10 @@ Examples {#examples}
   at some point script accesses a layout value in the element's subtree. In
   this situation, the user-agent cannot avoid rendering work and has to process
   any previously skipped rendering work in order to return a correct value to
-  the caller. Note that repeated calls to the same layout value should not
+  the caller. In this example, the result of getBoundingClientRect is a rect
+  positioned at (11, 22) with a size 100x200.
+
+  Note that repeated calls to the same layout value should not
   cause any additional rendering work, since the user-agent should retain the
   last updated rendering state.
 
--- a/index.bs
+++ b/index.bs
@@ -87,6 +87,37 @@ and pair that with a guarantee that when later rendered, it would not
 invalidate more than a small amount of style, layout, or paint in the rest of
 the document.
 
+Definitions {#definitions}
+===========
+
+* <dfn export>off-screen</dfn>: an element is considered to be off-screen if
+    its border box does not intersect the viewport with a fixed, user-agent
+    specified, margin. 
+    <div class=note>
+      The the user-agent is free to consider a fixed margin
+      around the viewport when computing whether the element is off-screen. This
+      means that when an element is approaching the viewport, the user-agent can
+      start processing it in preparation of the element entering the
+      viewport. A 50% viewport margin is a reasonable implementation default
+    </div>
+* <dfn export>on-screen</dfn>: an element is considered to be on-screen if it
+    is not [=off-screen=]
+* <dfn export>skipped</dfn>: an element is considered skipped if its subtree is
+    not painted or hit-tested. A skipped element has [=layout containment=],
+    [=paint containment=], and [=size containment=] applied to it in addition to
+    other containment values. The user-agent should avoid as much rendering work
+    in the skipped element's subtree as possible.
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1,317 +1,215 @@
...
 Work Status: exploring
-URL: https://tabatkins.github.io/specs/css-display-locking
+URL: https://wicg.github.io/display-locking
 Editor: Tab Atkins-Bittner, Google
-Abstract: This spec introduces features for "display locking",
-	allowing authors to more easily control when rendering work occurs,
-	making large amounts of content more performant.
+Editor: Vladimir Levin, Google
+Abstract: This spec introduces features for "display locking", allowing authors
+          to provide rendering performance hints to the user-agent. The user-agent can
```
chrishtr: They aren't really hints. knobs instead?

index.bs
```diff
@@ -1,317 +1,215 @@
...
-* Scrollers with a large amount of content,
-	without resorting to virtualization
-	(examples: Facebook and Twitter feeds,
-	CodeMirror documents)
+Render Subtree (a.k.a. Display Locking) is a CSS property designed to allow
+developers and browsers to easily scale to large amount of content and provide
+strong hints which allow the user-agent to delay rendering <a href=#f1>[1]</a>
+work.  More concretely, the goals are:
+
+* Avoid rendering work for content not visible to the user Support user-agent
```
chrishtr: I think you lost some formatting here.

index.bs
```diff
@@ -1,317 +1,215 @@
...
+* features and all layout algorithms (e.g. responsive design, flexbox, grid)
+    for this content
+
+Optimizing behavior and performance of the following use-cases motivate this work:
```
```suggestion
Optimize behavior and performance of the following use-cases motivate this work:
```

index.bs
```diff
@@ -1,317 +1,215 @@
...
-for more information on why it exists.
-
-Issue: TODO provide example of using this with render-subtree
-
-Issue: TODO provide example of using this for async-loaded content to avoid major layout shifts
-
+Note that in situations where 'render-subtree' values cause the element's
+subtree not to be painted or hit-tested, the user-agent should avoid as much
+rendering work as possible. As an example, because of the containment applied,
+in most cases it should be possible to skip determining layout box sizes and
```
```suggestion
in most cases it should be possible to skip computing styles, or determining layout box sizes and
```

index.bs
```diff
@@ -1,317 +1,215 @@
...
+positions unless explicitly requested by script.
```
```suggestion
positions, unless explicitly requested by script via APIs such as getBoundingClientRect.
```

index.bs
```diff
@@ -1,317 +1,215 @@
...
+    :: No effect; The element and its subtree are rendered as usual. This
+          provides no extra information or hints to the user-agent.
+
+    : <dfn export>hidden</dfn>
+    :: The element has [=layout containment=] and [=style containment=] applied
+         in addition to any existing containment values.
+
+         Additionally, if the element is <dfn export>off-screen</dfn>, in the
+         sense that it is not intersecting the viewport (with a reasonable
+         user-agent specified margin), then element has [=size containment=]
```
chrishtr: Suggest adding a definition called "hidden", which means:
* Size containment
* Not painted
* not hit-tested

Then you can say that when the element is off-screen, it is hidden; otherwise it is not hidden.

index.bs
```diff
@@ -1,317 +1,215 @@
...
+         applied in addition to any existing containment values and the
+         element's subtree is not painted or hit-tested. 
+
+    : <dfn export>hidden frozen</dfn>
+    :: The element has [=layout containment=], [=style containment=], and
+         [=size containment=] applied in addition to any existing containment
```
chrishtr: Substitute lines 117-119 with "hidden" according to above definition.

index.bs
```diff
@@ -1,317 +1,215 @@
...
-		<div class=locked>
-		  ... some content goes here ...
-		</div>
-	</xmp>
+  In this case, the element's subtree is not painted or hit-tested regardless
+  of viewport intersection. This means that the only way to have this subtree
+  painted is via script updating the value to remove render-subtree or change
+  its value. As before, the user-agent should skip as much of the rendering in
+  the subtree as possible. An additional effect of skipping rendering is that
+  the layout state of the subtree can be preserved, so that removing the
```
```suggestion
  the layout state of the subtree can be preserved by the user agent, so that removing the
```

index.bs
```diff
@@ -1,317 +1,215 @@
...
+  'render-subtree' property in the future will cause the subtree to be rendered
+  quicker than otherwise possible.
```
chrishtr: Also add an example where getBoundingClientRect is used together with render-subtree: hidden frozen

index.bs
vmpstr: I really don't know what to call it... I don't think they are knobs.. I think they are "very strong hints" Is there a better word than either hints or knobs?

index.bs
vmpstr: Oops.

index.bs
vmpstr: If you mean I should use the definition of 'hidden' in here, then I'm not sure how to do that since 'hidden' references on- and off- screen but this doesn't...

vmpstr:
PTAL.

index.bs
vmpstr: Oh i see what you mean above... See if what I have now is better, or whether it still makes sense to add a new definition.

index.bs
vmpstr: Oops missed this comment. Can you see if still makes sense in the new spec?

chrishtr: I think it would be better to first define "hidden" and "offscreen" explicitly, then reference them in the property definition. This way you don't have to repeat yourself for the "auto" vs "hidden" spec, and make it easier to read because "offscreen" has an intuitive meaning that everyone should guess at without having to read the exact definition.

chrishtr: "skipped" -> add on contain: size, not painted or hit tested [I named it "skipped" to differentiate from "hidden" which is a property value we're defining, as well as a thing elsewhere such as in visibility: hidden.

"offscreen" -> border box does not intersect the viewport plus fixed margin

"has containment" -> add on contain:style layout

Now the definitions:
"visible" -> no change
"auto": has containment, skipped when offscreen
"hidden": has containment, skipped.

vmpstr: Makes sense. Done.

vmpstr:
PTAL.

index.bs
```diff
@@ -1,317 +1,307 @@
...
+the developer to specify that specific parts of the DOM need not be rendered,
+and pair that with a guarantee that when later rendered, it would not
+invalidate more than a small amount of style, layout, or paint in the rest of
+the document.
+
+Definitions {#definitions}
+===========
+
+* <dfn export>off-screen</dfn>: an element is considered to be off-screen if
+    its border box does not intersect the viewport with a fixed, user-agent
```
```suggestion
    its border box does not intersect the visual viewport plus a fixed, user-agent
```

index.bs
```diff
@@ -1,317 +1,307 @@
...
+    specified, margin. 
+    <div class=note>
+      The the user-agent is free to consider a fixed margin
```
chrishtr: Maybe the UA should be allows to increase the margin to account for scroll velocity?

index.bs
```diff
@@ -1,317 +1,307 @@
...
 </dl>
 
-See also the <a href="https://github.com/WICG/display-locking/blob/master/explainer-contain-intrinsic-size.md">explainer for this property</a>
-for more information on why it exists.
-
-Issue: TODO provide example of using this with render-subtree
-
-Issue: TODO provide example of using this for async-loaded content to avoid major layout shifts
-
+When element is [=skipped=], it is possible to avoid parts of rendering in the
```
chrishtr: I think this paragraph duplicates sentences above?

index.bs
```diff
@@ -1,317 +1,307 @@
...
+element's subtree. In these situations, the user-agent should avoid as much
+rendering work as possible. This also means that the user-agent should retain
+the previously computed layout state if possible.
+
+<div class=note>
+  Intuitively, a [=hidden=] value means that the element behaves in a way that
+  does not expose the contents of its subtree to the user. Script interactions
```
```suggestion
  does not expose the contents of its subtree to the user (but still has an effect on layout, especially in conjunction with `contain-intrinsic-size`). Script interactions
```

index.bs
```diff
@@ -1,317 +1,307 @@
...
 		}
 		</style>
 
-		<div class=locked>
+		<div id=target class=sv>
 		  ... some content goes here ...
 		</div>
+    <script>
+      ...
+      target.firstElementChild.getBoundingClientRect();
```
chrishtr: Add a comment that this will force layout to occur if the UA had previously avoided it.

chrishtr: Also, add enough CSS details to the example that you can add to the comment the exact value that getBoundingClientRect() will return.

index.bs
vmpstr: Yeah I was thinking about this. I think it doesn't really make sense for us to put any arbitrary limits on this, but at the same time I don't want to focus the discussion on this margin. I'll just write that it's up to the user-agent to determine the margin

chrishtr: ok

vmpstr:
PTAL.

index.bs
```diff
@@ -91,14 +91,14 @@ Definitions {#definitions}
...
-      The the user-agent is free to consider a fixed margin
-      around the viewport when computing whether the element is off-screen. This
-      means that when an element is approaching the viewport, the user-agent can
-      start processing it in preparation of the element entering the
-      viewport. A 50% viewport margin is a reasonable implementation default
+      The the user-agent is free to consider any margin around the viewport
+      when computing whether the element is off-screen. The margin is meant to
+      allow the user-agent to start preparing the element ahead of time, before
+      it enters the viewport. A 50% viewport margin is a reasonable
+      implementation default
```
```suggestion
      implementation default.
```

PR #124, opened 2020-02-27
Added accessibility section

vmpstr:
Please take a look. Rendered output: https://vmpstr.github.io/display-locking/#accessibility

/cc @alice

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -184,7 +184,7 @@ tree</a> and <a href="https://w3c.github.io/css-aam/#dfn-assistive-technology">
 assistive technologies</a>:
 * Subtrees affected by ''subtree-visibility: visible'' or ''subtree-visibility: auto''
     that are not [=skipped=] should be included in the accessibility tree as usual.
-* Subtrees of [=skipped=] elements not be included in the accessbility tree,
+* Subtrees of [=skipped=] elements should not be included in the accessibility tree,
     with the following exception:
     * Subtrees which are affected by ''subtree-visibility: auto'' and are
         [=skipped=] should be also be included in the accessibility tree
--- a/index.bs
+++ b/index.bs
@@ -183,7 +183,7 @@ subtree, it also affects information exposed to the
 tree</a> and <a href="https://w3c.github.io/css-aam/#dfn-assistive-technology">
 assistive technologies</a>:
 * Subtrees affected by ''subtree-visibility: visible'' or ''subtree-visibility: auto''
-    and they are not [=skipped=] should be included in the accessibility tree as usual.
+    that are not [=skipped=] should be included in the accessibility tree as usual.
 * Subtrees of [=skipped=] elements not be included in the accessbility tree,
     with the following exception:
     * Subtrees which are affected by ''subtree-visibility: auto'' and are
--- a/index.bs
+++ b/index.bs
@@ -182,15 +182,23 @@ subtree, it also affects information exposed to the
 <a href="https://w3c.github.io/css-aam/#dfn-accessibility-tree">accessibility
 tree</a> and <a href="https://w3c.github.io/css-aam/#dfn-assistive-technology">
 assistive technologies</a>:
-* The structure of the page should be exposed to assistive technologies.
-* When [=skipped=], the element's subtree should be exposed as
-    <a href="https://w3c.github.io/css-aam/#dfn-hidden">hidden</a>.
+* Subtrees affected by ''subtree-visibility: visible'' or ''subtree-visibility: auto''
+    and they are not [=skipped=] should be included in the accessibility tree as usual.
+* Subtrees of [=skipped=] elements not be included in the accessbility tree,
+    with the following exception:
+    * Subtrees which are affected by ''subtree-visibility: auto'' and are
+        [=skipped=] should be also be included in the accessibility tree
+        subject to the rendering constraints below.
 
 <div class=note>
-  The general idea for exposing elements as hidden is the following: if content
-  is not observable in the painted output of the page, then it should not be
-  observable by assistive technologies user. Similarly, if the content is
-  painted then it should be available to the assistive technologies user.
+  Since assistive technologies may provide quick access to offscreen elements,
+  it is desirable that elements in ''subtree-visibility: auto'' subtrees be made
+  available to assistive technologies to provide this functionality, since they
+  are intended to be observable to users.
+
+  Conversely, since elements in ''subtree-visibility: hidden'' subtrees are not
+  intended to be observable by users, they should not be exposed to assistive
+  technology.
 </div>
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -165,6 +165,42 @@ the user-agent should retain the previously computed layout state if possible.
...
 
+Accessibility {#accessibility}
+=============
+
+Similar to the way 'subtree-visibility' affects painted output of the element's
+subtree, it also affects information exposed to the
+<a href="https://w3c.github.io/css-aam/#dfn-accessibility-tree">accessibility
+tree</a> and <a href="https://w3c.github.io/css-aam/#dfn-assistive-technology">
+assistive technologies</a>:
+* The structure of the page should be exposed to assistive technologies.
```
alice: Perhaps flesh this out a little more, something like

* Subtrees with a `subtree-visibility` value of `auto` or `visible` should be included in the accessibility tree as usual.
* Subtrees with a `subtree-visibility` value of `skipped` should be excluded from the accessibility tree.

index.bs
```diff
@@ -165,6 +165,42 @@ the user-agent should retain the previously computed layout state if possible.
...
+* When [=skipped=], the element's subtree should be exposed as
+    <a href="https://w3c.github.io/css-aam/#dfn-hidden">hidden</a>.
+
+<div class=note>
+  The general idea for exposing elements as hidden is the following: if content
```
alice: Suggested rephrasing:

Since assistive technologies may provide quick access to offscreen elements, it is important that elements in `subtree-visibility: auto` subtrees be made available to assistive technologies to provide this functionality, since they are intended to be observable to users.

Conversely, since elements in `subtree-visibility: hidden` subtrees are not intended to be observable by users, they should not be exposed to assistive technology.

index.bs
```diff
@@ -165,6 +165,42 @@ the user-agent should retain the previously computed layout state if possible.
...
+
+<div class=note>
+  The requirement of rendering performance equivalency stems from privacy
+  considerations. It is imperative that the user-agent ensures that a page
+  using 'subtree-visibility' cannot use timing information to deduce whether
+  the user is using assistive technologies. For this reason, the rendering
+  performance of information exposed to assistive technologies must be the same
+  as the rendering performance of information exposed to painted output.
+
+  In other words, if the user-agent omits rendering work when producing painted
```
alice: This is nicely worded.

index.bs
vmpstr: This particular phrasing seems to contradict the normative text. The problem is specifically about the "auto" value. When off-screen, the elements are not rendered (ie they are 'skipped'). So they should be excluded from the accessibility tree (according to the bullet points above).

Here, however, we are saying that should be included for quick off-screen access. I think what's missing is the distinction between exposing the element vs exposing the layout information (size and positioning). 

I think we should expose the element (which is what I meant by the 'structure of the page'). But we can't expose the layout information, since that would cause rendering to me noticeably slower. 

Do you know if there's terminology here that we can use to convey this?

alice: > When off-screen, the elements are not rendered (ie they are 'skipped'). So they should be excluded from the accessibility tree (according to the bullet points above).

Which bullet points?

> I think we should expose the element (which is what I meant by the 'structure of the page'). 

Hm, I found that wording hard to follow. It's ambiguous, I think.

> But we can't expose the layout information, since that would cause rendering to me noticeably slower.

Agreed.

> Do you know if there's terminology here that we can use to convey this?

I thought it was covered by the paragraph below.

vmpstr: > > When off-screen, the elements are not rendered (ie they are 'skipped'). So they should be excluded from the accessibility tree (according to the bullet points above).
> 
> Which bullet points?

Sorry, I was referencing the bullet points you suggested in the previous comment. Specifically the second one says:

* Subtrees with a subtree-visibility value of skipped should be excluded from the accessibility tree.

"skipped" as we defined it in the spec means not-rendered: this would be, among others, the off-screen elements with `subtree-visibility: auto`. So, that bullet point says don't include off-screen elements with `subtree-visibility: auto` in the accessibility tree. 

However, the suggested phrasing of this paragraph says that off-screen elements with `subtree-visibility: auto` should be made available to the assistive technologies.

Doesn't that mean we should include them in the accessibility tree? Or is there a distinction between including something in the accessibility tree vs making it available to assistive technology?

index.bs
vmpstr: Thanks! :)

index.bs

index.bs
vmpstr: I've rephrased this a bit, new text: https://vmpstr.github.io/display-locking/#accessibility 

Let me know if that reads ok, or whether I should make more changes. I'd rather get this text correct and have it be unambiguous

index.bs
```diff
@@ -165,6 +165,60 @@ the user-agent should retain the previously computed layout state if possible.
...
+Accessibility {#accessibility}
+=============
+
+Similar to the way 'subtree-visibility' affects painted output of the element's
+subtree, it also affects information exposed to the
+<a href="https://w3c.github.io/css-aam/#dfn-accessibility-tree">accessibility
+tree</a> and <a href="https://w3c.github.io/css-aam/#dfn-assistive-technology">
+assistive technologies</a>:
+* Subtrees affected by ''subtree-visibility: visible'' or ''subtree-visibility: auto''
+    and they are not [=skipped=] should be included in the accessibility tree as usual.
```
```suggestion
    that are not [=skipped=] should be included in the accessibility tree as usual.
```

index.bs
```diff
@@ -165,6 +165,60 @@ the user-agent should retain the previously computed layout state if possible.
...
+* Subtrees of [=skipped=] elements not be included in the accessbility tree,
```
```suggestion
* Subtrees of [=skipped=] elements should not be included in the accessibility tree,
```

index.bs
```diff
@@ -165,6 +165,60 @@ the user-agent should retain the previously computed layout state if possible.
...
+  The requirement of rendering performance equivalency stems from privacy
+  considerations. It is imperative that the user-agent ensures that a page
+  using 'subtree-visibility' cannot use timing information to deduce whether
+  the user is using assistive technologies. For this reason, the rendering
+  performance of information exposed to assistive technologies must be the same
+  as the rendering performance of information exposed to painted output.
+
+  In other words, if the user-agent omits rendering work when producing painted
+  output, then it must also omit the same rendering work when providing
+  information for assistive technologies.
```
```suggestion
  information for assistive technologies, and vice-versa.
```

vmpstr:
@alice could you please take another look when you get a chance?

index.bs
chrishtr: I think the new phrasing looks great! Let's commit this PR. Alice, if you have more comments please add them and we can add them in a new PR.

vmpstr:
Committed per chrishtr's comment; we can adjust in another PR if needed.

PR #136, opened 2020-03-09
Add select and focus to subtree-visibility spec

vmpstr:
Please take a look: https://vmpstr.github.io/display-locking/#subtree-visibility

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -179,8 +179,8 @@ the user-agent should retain the previously computed layout state if possible.
   The fact that [=off-screen=] elements with ''subtree-visibility: auto'' are
   [=skipped=] is a rendering performance optimization.
 
-  Note that selected and focused elements affected by ''subtree-visibility:
-  auto'' are treated as if they are painted, since users can interact with this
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -138,7 +138,16 @@ Inherited: no
     :: No effect. This provides no extra information or hints to the user-agent.
 
     : <dfn export>auto</dfn>
-    :: The element [=gains containment=]. If the element is [=off-screen=], it is [=skipped=].
+    :: The element [=gains containment=]. 
+         If the following conditions hold then the element is also [=skipped=]:
```
```suggestion
         If all the following conditions hold then the element is also [=skipped=]:
```

index.bs
```diff
@@ -138,7 +138,16 @@ Inherited: no
...
+           * The element is [=off-screen=].
+           * Neither the element nor any of its [=subtree=] elements are
```
chrishtr: Can you link to the HTML spec concept for focus?

vmpstr:
PTAL

PR #137, opened 2020-03-16
Added restrictions and clarifications

### LINKED ISSUE #81: Spec: if containment can't apply then s-v has no effect (e.g. display: contents)
I don't think we can reasonably expect to lock a display: contents element, since it has no box for which children can be skipped. That being said, in the imperative version we've relied on promise rejection or displayLock.locked value to figure out whether we're actually locked.

What should we do in the declarative version? Should we just clear the rendersubtree attribute? Or do nothing?

/cc @chrishtr @rakina

vmpstr:
https://vmpstr.github.io/display-locking/index.html#restrictions

Closes #81 

Please take a look.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -203,7 +203,7 @@ Restrictions and Clarifications {#restrictions}
     have no effect.
 
 * <a href="https://html.spec.whatwg.org/multipage/rendering.html#replaced-elements">
-    Replaced elements</a>, such as images, do not paint their contents if they
+    Replaced elements</a> do not paint their contents if they
     are [=skipped=] due to 'subtree-visibility'. That is, the element's border
     and background are painted, but the replaced content -- as described in
     steps 7.1 and 7.2.4 of <a
--- a/index.bs
+++ b/index.bs
@@ -195,6 +195,19 @@ the user-agent should retain the previously computed layout state if possible.
   values are added on top of any existing containment values.
 </div>
 
+Restrictions and Clarifications {#restrictions}
+===============================
+
+* In situations where [=layout containment=] has no effect (e.g. the
+    element does not generate a principal box), 'subtree-visibility' values also
+    have no effect. In other words, if [=layout containment=] has no effect, the
+    'subtree-visibility' value is effectively [=visible=].
+
+* <a href="https://html.spec.whatwg.org/multipage/rendering.html#replaced-elements">
+    Replaced elements</a>, such as images, do not paint their contents if they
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -195,6 +195,19 @@ the user-agent should retain the previously computed layout state if possible.
   values are added on top of any existing containment values.
 </div>
 
+Restrictions and Clarifications {#restrictions}
+===============================
+
+* In situations where [=layout containment=] has no effect (e.g. the
+    element does not generate a principal box), 'subtree-visibility' values also
+    have no effect. In other words, if [=layout containment=] has no effect, the
```
chrishtr: Consider removing the second sentence, seems redundant.

index.bs
```diff
@@ -195,6 +195,19 @@ the user-agent should retain the previously computed layout state if possible.
...
+    'subtree-visibility' value is effectively [=visible=].
+
+* <a href="https://html.spec.whatwg.org/multipage/rendering.html#replaced-elements">
```
chrishtr: What about:

Object, embed, iframe?

index.bs

index.bs
vmpstr: We also don't paint iframe contents (we have tests for that). I'll add tests for object and embed, but I think we should not separate things into different subgroups of replaced if we don't have a reason to do so. So, I think this phrasing _should_ be correct. WDYT?

chrishtr: Ok, but then you should either omit "such as images" to avoid it sounding like it's just for images.

index.bs
```diff
@@ -195,6 +195,18 @@ the user-agent should retain the previously computed layout state if possible.
...
 
+Restrictions and Clarifications {#restrictions}
+===============================
+
+* In situations where [=layout containment=] has no effect (e.g. the
+    element does not generate a principal box), 'subtree-visibility' values also
+    have no effect.
+
+* <a href="https://html.spec.whatwg.org/multipage/rendering.html#replaced-elements">
+    Replaced elements</a>, such as images, do not paint their contents if they
```
chrishtr: Suggest further specifying that "do not paint their contents" means not running the parts of step 7.1 or step 7.4 from this algorithm that refer to replaced contents:

https://www.w3.org/TR/CSS21/zindex.html

vmpstr: Done. Let me know if this reads ok.

vmpstr:
ptal

PR #138, opened 2020-03-17
Added intersection and resize observer points

### LINKED ISSUE #71: Spec ResizeObserver and IntersectionObserver interactions.
2 approaches seem possible

1 when invisible, observers stop triggering, when the observed element becomes visible again, it's size is compared to it's previously visible size.
2 when invisible, observers continue to trigger and anything which results in a layout of the observed element may trigger the observer

Number 1 reduces spurious work but probably requires us to save the pre-invisible size somewhere (and I don't think that somewhere currently exists).
Number 2 is simpler but could cause expensive work to be done pointlessly on invisible elements.

vmpstr:
Closes #71

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -221,7 +221,7 @@ Restrictions and Clarifications {#restrictions}
     href="https://drafts.csswg.org/resize-observer/#resize-observer-interface">resize
     observer</a>, elements in the subtree of a [=skipped=] element do not
     change their size. If these elements become painted again, the resize
-    observation will be delivered if the new painted size differs frrom the
+    observation will be delivered if the new painted size differs from the
     last size used to notify the resize observer.
 
 Accessibility {#accessibility}
--- a/index.bs
+++ b/index.bs
@@ -210,6 +210,20 @@ Restrictions and Clarifications {#restrictions}
     href="https://www.w3.org/TR/CSS21/zindex.html#painting-order">the painting
     order</a> steps -- is not.
 
+* From the perspective of an <a
+    href="https://w3c.github.io/IntersectionObserver/">intersection observer</a>,
+    elements in the subtree of a [=skipped=] element are not intersecting the <a
+    href="https://w3c.github.io/IntersectionObserver/#dom-intersectionobserver-root">root</a>.
+    This is true even if both the root and the target elements are in the subtree
+    of a [=skipped=] element.
+
+* From the perpsective of a <a
+    href="https://drafts.csswg.org/resize-observer/#resize-observer-interface">resize
+    observer</a>, elements in the subtree of a [=skipped=] element do not
+    change their size. If these elements become painted again, the resize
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -210,6 +210,20 @@ Restrictions and Clarifications {#restrictions}
     href="https://www.w3.org/TR/CSS21/zindex.html#painting-order">the painting
     order</a> steps -- is not.
 
+* From the perspective of an <a
+    href="https://w3c.github.io/IntersectionObserver/">intersection observer</a>,
+    elements in the subtree of a [=skipped=] element are not intersecting the <a
+    href="https://w3c.github.io/IntersectionObserver/#dom-intersectionobserver-root">root</a>.
+    This is true even if both the root and the target elements are in the subtree
```
vmpstr: I think this makes sense, since the target and the root are invisible in this case

chrishtr: Use the "subtree" definition in this spec here and elsewhere to make it clear you mean flat tree.

index.bs
```diff
@@ -210,6 +210,20 @@ Restrictions and Clarifications {#restrictions}
...
+    of a [=skipped=] element.
+
+* From the perpsective of a <a
+    href="https://drafts.csswg.org/resize-observer/#resize-observer-interface">resize
+    observer</a>, elements in the subtree of a [=skipped=] element do not
+    change their size. If these elements become painted again, the resize
+    observation will be delivered if the new painted size differs from the
+    last size used to notify the resize observer.
```
vmpstr: This last part may be contentious... One could argue that because the element was hidden it was of an unknown size, so when it becomes painted again we should always deliver a resize observation. I'm not really sure. WDYT?

index.bs
```diff
@@ -210,6 +210,20 @@ Restrictions and Clarifications {#restrictions}
...
+    href="https://w3c.github.io/IntersectionObserver/">intersection observer</a>,
+    elements in the subtree of a [=skipped=] element are not intersecting the <a
+    href="https://w3c.github.io/IntersectionObserver/#dom-intersectionobserver-root">root</a>.
+    This is true even if both the root and the target elements are in the subtree
+    of a [=skipped=] element.
+
+* From the perpsective of a <a
+    href="https://drafts.csswg.org/resize-observer/#resize-observer-interface">resize
+    observer</a>, elements in the subtree of a [=skipped=] element do not
+    change their size. If these elements become painted again, the resize
```
```suggestion
    change their size. If these elements become non-skipped again, the resize
```

index.bs
```diff
@@ -210,6 +210,20 @@ Restrictions and Clarifications {#restrictions}
...
+    observation will be delivered if the new painted size differs frrom the
```
```suggestion
    observation will be delivered if the new size differs from the
```

index.bs

vmpstr:
PTAL

PR #142, opened 2020-03-21
beforematch security and privacy self-review

josepharhar:


marcoscaceres:
Hi @josepharhar. This looks great. Please be sure to join the WICG so we can merge this. You should have gotten an email. Thanks :)

marcoscaceres:
(just in case you've already joined the WICG, you can head over to the following to connect your W3C and GitHub account: https://www.w3.org/users/myprofile/connectedaccounts )

beforematch-security-and-privacy-assessment.md
```diff
@@ -0,0 +1,167 @@
+# beforematch event security and privacy self-review questionnaire
+
+The beforematch event ([explainer](https://github.com/WICG/display-locking/blob/master/explainer-beforematch.md)) is fired on elements before scrolling occurs containing matching text in these three cases:
```
vmpstr:
```suggestion
The `beforematch` event ([explainer](https://github.com/WICG/display-locking/blob/master/explainer-beforematch.md)) is fired on elements in the following cases, before the scrolling occurs:
```

beforematch-security-and-privacy-assessment.md
```diff
@@ -0,0 +1,167 @@
...
+
+### 2.1. What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
+
+This feature may expose hints as to what the user is searching for to webpages. I wrote more detailed considerations about this in the "Privacy Considerations" section at the bottom of this document.
+
+This exposure is necessary in order to fulfill the use case of displaying content that was searched for, as described in [the beforematch explainer](https://github.com/WICG/display-locking/blob/master/explainer-beforematch.md).
+
+### 2.2. Is this specification exposing the minimum amount of information necessary to power the feature?
+
+For the use case of modifying the render-subtree css property on matching elements, less information would be exposed if we just automatically changed the property to expand the search result when searching instead. Although you could use the render-subtree property and a mutation observer to see where matches occur to find out which elements matched what the user searched for, this is more restrictive than firing the beforematch event every time a search occurs regardless of the render-subtree property.
```
vmpstr: s/render-subtree/subtree-visibility/

So, I think you're answering the question kind of holistically here considering the whole subtree-visibility / display locking project. That is, the argument is that we could expose less information if we didn't have the `beforematch` event at all, which is true. However, when answering the question "are we exposing the minimum amount of information _to make `beforematch` possible_", I think the answer would be different. WDYT?

To put this differently, I think we can discuss whether `beforematch` is the right event to expose separately from discussing the privacy and security implications of `beforematch`; and this document is meant to do the latter.

beforematch-security-and-privacy-assessment.md
```diff
@@ -0,0 +1,167 @@
...
+
+This feature allows the page to respond to find-in-page, which is part of the user agent's native UI. However, this feature does not allow control over find-in-page.
+
+### 2.12. What temporary identifiers might this specification create or expose to the web?
+
+None.
+
+### 2.13. How does this specification distinguish between behavior in first-party and third-party contexts?
+
+I'm assuming first-party and third-party contexts here refers to cross origin iframes.
```
vmpstr: Yeah, I think you're right.

beforematch-security-and-privacy-assessment.md
```diff
@@ -0,0 +1,167 @@
...
+* In the ElementFragment case, the scrolling, fragments, and beforematch events are fully insulated from each other when they occur in different cross origin iframes.
+* In the ScrollToTextFragment case, text inside of a cross origin iframe will not be searched for or scrolled to. Therefore, no beforematch event will be emitted.
+* In the find-in-page case, content in cross origin iframes can be searched for and scrolled to. I'm not sure if and where we should fire beforematch in this case.
```
vmpstr: The browser would coordinate searching across different frames, so the event should only fire within the frame that contains a match

beforematch-security-and-privacy-assessment.md
```diff
@@ -0,0 +1,167 @@
...
+
+### 2.14. How does this specification work in the context of a user agent's Private Browsing or "incognito" mode?
+
+This feature will work the same regardless of incognito mode being enabled or disabled.
+
+### 2.15. Does this specification have a "Security Considerations" and "Privacy Considerations" section?
+
+I included a Privacy Considerations section further down this document. I did not include a Security Considerations section because I do not believe this feature exposes any security vulnerabilities.
```
vmpstr: FWIW, when we write the spec for this feature, I think we should include a Privacy section in there

beforematch-security-and-privacy-assessment.md
josepharhar: I see, you have a good point. Without thinking about subtree-visibility, I'm not really sure what there is to talk about here. Is there another use case I could or should think from? Should I just change the wording?

beforematch-security-and-privacy-assessment.md
josepharhar: Should I keep this one? I'd be hesitant to remove it and say it will just be included later in a spec...

beforematch-security-and-privacy-assessment.md
josepharhar: Right, that makes perfect sense and doesn't really require any additional implementation details. I'll clarify it.

beforematch-security-and-privacy-assessment.md
josepharhar: Cool, I'll reword this as more of an assertion and less of a question.

josepharhar:
> (just in case you've already joined the WICG, you can head over to the following to connect your W3C and GitHub account: https://www.w3.org/users/myprofile/connectedaccounts )

Thanks for the link, I have connected my github account.

beforematch-security-and-privacy-assessment.md
vmpstr: I think this should just say something like

"""
Yes, this is exposing only the information required for the `beforematch` event to function as specified.
"""

In other words, It doesn't leak anything extra information, it just exposes the information that it is meant to expose by design.

beforematch-security-and-privacy-assessment.md
vmpstr: You can keep this one. It's just this questionnaire is usually done when there is already a spec draft, and it's asking whether the spec draft as this section :) When we have the spec draft, we definitely will have this section, but it's good to keep it here for now as well.

josepharhar: Sounds good, thanks for letting me know

beforematch-security-and-privacy-assessment.md
josepharhar: Sounds good

beforematch-security-and-privacy-assessment.md
```diff
@@ -0,0 +1,165 @@
...
+  const elementFromFragment = document.getElementById(window.location.hash);
+});
+```
+* In the ScrollToTextFragment case, the beforematch event will hint to what the user was searching for by firing on the block level element containing the text which was searched for. However, the :target pseudo selector which is already added by ScrollToTextFragment is also applied to the same element that beforematch would fire on, so the beforematch event will not expose anything new in this case.
+* In the find-in-page case, the same information will be exposed as the ScrollToTextFragment case. However, if the user is incrementally searching by typing in a search character by character, it will result in repeated beforematch events being fired. It may be possible to reconstruct what the user is searching for in this case by having elements in the page which match every next possible character the user could search for, and then when a match occurs, changing the text of all of the elements to contain every next possible match, and so on. Here is an example page which conveys this idea. I couldn't actually get it to work properly, but with some more work I think it might:
```
vmpstr, line 100:
```suggestion
* In the find-in-page case, the same information will be exposed as the ScrollToTextFragment case. However, if the user is incrementally searching by typing in a search dialog character by character, it will result in repeated beforematch events being fired. It may be possible to reconstruct what the user is searching for in this case by having elements in the page which match every next possible character the user could search for, and then when a match occurs, changing the text of all of the elements to contain every next possible match, and so on. Here is an example page which conveys this idea. I couldn't actually get it to work properly, but with some more work I think it might:
```

beforematch-security-and-privacy-assessment.md
```diff
@@ -0,0 +1,165 @@
...
+
+However, find-in-page attacks may already be possible due to the fact that pages can already intercept ctrl+f and already have the power to spoof the functionality and appearance of the browser's native find-in-page. In fact, doing this may be even easier and more reliable than using beforematch to extract the user's search text.
+
+For the ScrollToTextFragment case, you can already figure out what block-level element containing the search result was by looking at the element with the :target pseudo selector.
+
+### 3. What privacy mitigations have been implemented?
+
+No privacy mitigations have been implemented.
+
+For the find-in-page privacy flaw, I could imagine a mitigation where we would choose not to fire beforematch repeatedly in this case. In the example page above, the scroll offset would not be change in between firings of the event, which we could use as a heuristic to combat this case. However, I'm not sure that it couldn't be circumvented or if it would cause issues when trying to use the feature legitimately.
```
vmpstr, line 159:
```suggestion
For the find-in-page privacy case, I could imagine a mitigation where we would choose not to fire beforematch repeatedly in this case. In the example page above, the scroll offset would not be change in between firings of the event, which we could use as a heuristic to combat this case. However, I'm not sure that it couldn't be circumvented or if it would cause issues when trying to use the feature legitimately.
```

PR #148, opened 2020-03-31
Added iframe clarification

### LINKED ISSUE #146: add clarifications to subtree-visibility spec about iframes
1. All elements inside iframes under a skipped `<iframe>` element are skipped.
2. The "update the rendering" steps for an iframe under a skipped `<iframe>` element is not run.

vmpstr:
Closes #146

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -286,10 +286,11 @@ Restrictions and Clarifications {#restrictions}
 
 8. If an <a
     href="https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element">iframe</a>
-    element is [=skipped=], then the content of the document inside the iframe does not paint or
-    participate in hit testing. The user-agent should skip <a
+    element is [=skipped=], then the content of the document inside the iframe
+    does not paint or participate in hit testing. The user-agent should skip <a
     href="https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering">Update
-    the Rendering</a> step of the Processing Model in the frame.
+    the Rendering</a> step of the Processing Model in the frame's event loop,
+    if possible (see note below).
     <div class=note>
       As is the case with other elements, contents of an iframe are not
       painted. With iframes there is an opportunity to skip all of the steps of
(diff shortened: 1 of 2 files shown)


vmpstr:
PTAL

index.bs
```diff
@@ -284,6 +284,22 @@ Restrictions and Clarifications {#restrictions}
       function and user gestures.
     </note>
 
+8. If an <a
+    href="https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element">iframe</a>
+    element is [=skipped=], then the content of the frame does not paint or
```
```suggestion
    element is [=skipped=], then the content of the document inside the iframe does not paint or
```

index.bs
```diff
@@ -284,6 +284,22 @@ Restrictions and Clarifications {#restrictions}
...
+    participate in hit testing. The user-agent should skip <a
+    href="https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering">Update
+    the Rendering</a> step of the Processing Model in the frame, if possible.
```
```suggestion
    the Rendering</a> step of the Processing Model in the frame.
```

chrishtr: Also, s/frame/frame's window's event loop/

index.bs
```diff
@@ -284,6 +284,22 @@ Restrictions and Clarifications {#restrictions}
...
+    <div class=note>
+      As is the case with other elements, contents of an iframe are not
+      painted. With iframes there is an opportunity to skip all of the steps of
+      the Update the Rendering step. The user-agent should skip this work if
```
```suggestion
      the Update the Rendering steps. The user-agent should skip this work if
```

index.bs
chrishtr: Re "if possible": I see what you mean now, based on the note below...

index.bs
chrishtr: And all descendant documents in the frame tree

index.bs
vmpstr: I reworded this a bit. frame's window's event loop is extremely confusing to parse... I said frame's event loop, which I think captures sufficient information. WDYT?

Also added "see note below" after "if possible"

vmpstr:
PTAL

PR #149, opened 2020-04-01
Clarify innerText behavior.

### LINKED ISSUE #54: Spec innerText interactions with s-v
I think we should not yield any innerText results when querying a locked subtree (similar to visibility: hidden).

vmpstr:
Closes #54

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -301,7 +301,7 @@ Restrictions and Clarifications {#restrictions}
       step has to run at least once.
     </div>
 
-9. Neither the [=skipped=] element nor any of its subtree elements contribute
+9. The subtree of a [=skipped=] element does not contribute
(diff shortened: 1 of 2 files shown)


vmpstr:
PTAL

index.bs
```diff
@@ -301,6 +301,10 @@ Restrictions and Clarifications {#restrictions}
       step has to run at least once.
     </div>
 
+9. Neither the [=skipped=] element nor any of its subtree elements contribute
```
chrishtr, line 304:
```suggestion
9. The subtree of a [=skipped=] element does not contribute
```

vmpstr:
Yeah that absolutely makes sense.

vmpstr:
PTAL

PR #151, opened 2020-04-07
added paint containment

### LINKED ISSUE #147: [css-subtree-visibility-1] definition of 'subtree-visibility: auto' mandates flickering of ink overflowing descendants when scrolling
I believe the current definition of `subtree-visibility: auto` mandates flickering of ink overflowing descendants when scrolling the page.

In particular, the current [definition of `auto`](https://wicg.github.io/display-locking/index.html#subtree-visibility-auto) refers to a [definition of "off-screen"](https://wicg.github.io/display-locking/index.html#off-screen) that defines being off-screen in terms of the element's border box.

This means that when the element's border box is offscreen (and outside whatever the UA-specific margin is) but some of it descendants are still onscreen, then those descendants are *required* not to be painted.

This means that the spec requires that, when scrolling, those descendants flicker visible/invisible as a function of the border-box's position in the viewport.

This seems very bad from a user interface perspective (visible flickering) and also probably bad from a performance perspective (mandating repainting during scroll).

It seems like it would be desirable for these elements to be forced to have `contain: paint` (or at least the overflow-related aspects of `contain: paint`; the z-ordering aspects of it don't seem to be needed, and `contain: paint` could perhaps be divided in half).

(I realize, as @chrishtr pointed out to me, that changing the "border box" definition isn't possible, because the point of the feature is to allow deferring of layout work inside the subtree until it's close to being onscreen.)

vmpstr:
Closes #147

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -113,9 +113,10 @@ Definitions {#definitions}
     is not [=off-screen=]
 * <dfn export>skipped</dfn>: an element is considered skipped if its [=subtree=] is
     not painted or hit-tested. A skipped element has [=layout containment=],
-    [=style containment=], and [=size containment=] applied to it in addition to
-    other containment values. The user-agent should avoid as much rendering work
-    in the skipped element's [=subtree=] as possible.
+    [=style containment=], [=paint containment=], and [=size containment=]
+    applied to it in addition to other containment values. The user-agent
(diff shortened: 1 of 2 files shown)


vmpstr:
Please take a look.

cbiesinger:
Looks reasonable to me!

PR #152, opened 2020-04-08
Clean up the subtree-visibility spec language.

vmpstr:
This PR cleans up some spec language:
* Applies suggestions from https://github.com/w3c/csswg-drafts/issues/4843
* Replaces some references to "painted" with "non-skipped"
* Formats the index.bs file


vmpstr:
PTAL

PR #153, opened 2020-04-16
Added privacy & security section

vmpstr:


vmpstr:
Please take a look.

vmpstr:
@cbiesinger could you please take a look as well?

index.bs
```diff
@@ -385,6 +385,27 @@ assistive technologies</a>:
...
+of using this particular feature for accessibility detection.
+
+Other than the use case for detecting accessibility, this feature does not deal
+with any personally identifiable or sensitive information. Hence, the risk to
+the user is low.
+
+When it comes to dealing with cross-origin frames, the usage of this feature
+does not enable any new detections. Frames are isolated in terms of
+performance, so user-agents are capable of optimizing the frame performance
+without this feature, in the same ways that this feature recommends.
```
chrishtr, line 407: A cross-origin iframe can learn that it is offscreen, and under a subtree-visibility hidden or auto ancestor, by noticing that requestAnimationFrame callbacks are not run. However, IntersectionObserver can already directly tell such iframes whether they are offscreen, and browsers are already allowed to throttle or stop generating frames for off-screen cross-origin iframes (reference to HTML spec update-the-rendering).

vmpstr: I've reworded this to incorporate your comment. However, it's not a word for word copy. The problem with your phrasing is that it seems to say
"iframe can detect subtree-visibility. however, it can't". 
I (tried to) rephrased it more to say
"it may seem like it can, but it can't because..."

Issue #150, opened 2020-04-02
beforematch: what to do when the element is no longer matching after callback.


vmpstr:
Beforematch puts in a script callback before we scroll to whatever caused us to fire the signal (scroll-to-text, find-in-page, or fragment link navigation).

The question is what should we do if the conditions that caused us to find the match no longer hold after beforematch runs.

For example, in find-in-page case, the callback can:
* Add display: none to some ancestor
* Change text
* Move the node before/after other matches

We need to know how to handle all of those. Hopefully, whatever the decision, we can apply to other situations that we might not have thought of yet. 

/cc @josepharhar @chrishtr @domenic

josepharhar:
Some thoughts:
- find-in-page, ScrollToTextFragment, and ElementFragment each have their way of deciding to match something or not, maybe it would make sense to run that a second time after the beforematch event, perhaps based on what changes in css properties we see.
- If an element is display locked, we should fire the beforematch event on it. If the displaylock isn't removed, then it is an important for that use case that we _don't_ scroll to it, right? In find-in-page, ScrollToTextFragment, and ElementFragment is it important to run another search again in this case?

josepharhar:
I documented a bunch of current scrolling behavior here: https://github.com/WICG/display-locking/pull/154

Another thought brought up in the PR: When we choose to scroll to a different element due to DOM/style changes in the beforematch handler, should said different element also get a beforematch event?
My gut feeling is yes, we should fire two events. At the moment, ScrollToTextFragment doesn't fire a second beforematch event when the target element changes due to the original target element being removed from the DOM.

josepharhar:
Some more thoughts from @vmpstr: We should continue to fire events when the element we are trying to scroll to is changing, but we should limit it to like 10 to prevent infinite loops

PR #154, opened 2020-04-23
Add beforematch scrolling behavior

josepharhar:
I made this to address this code review comment: https://chromium-review.googlesource.com/c/chromium/src/+/2068873/21#message-93c7132faae9655a9a22e6b38ed73ff90315d6cb

### DIFF
--- a/explainer-beforematch.md
+++ b/explainer-beforematch.md
@@ -191,6 +191,91 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
 However, the privacy aspect of the event should be discussed in more detail via
 formal security reviews.
 
+### Responses to DOM and style changes in the `beforematch` event handler
+
+The scrolling behavior exhibited by find-in-page, ScrollToTextFragment, and
+ElementFragment after style and DOM changes are made by `beforematch` event
+handlers is not well specified yet and is being discussed here: 
+https://github.com/WICG/display-locking/issues/150
+
+Here is a list of `beforematch` handler use cases that could affect scrolling:
+1. **Removed from DOM**: The `beforematch` event handler removes the
+target element from the DOM.
+2. **Reparented in DOM**: The `beforematch` event handler removes the target
+element from the DOM, and re-adds it further down the tree such that the
+location to scroll to is different.
+3. **display: none**: The `beforematch` event handler adds the style
+`display: none` to the target element.
+4. **visibility: hidden**: The `beforematch` event handler adds the style
+`visibility: hidden` to the target element.
+5. **subtree-visibility: hidden-matchable -> visible**: The target element has
+the style `subtree-visibility: hidden-matchable` before the `beforematch` event
+is fired on the target element, and the `beforematch` event handler changes the
+style value from `hidden-matchable` to `visible`.
+
+
+Here is the current scrolling behavior in Chromium for each of those cases in
+ScrollToTextFragment, and ElementFragment, and find-in-page. As mentioned
+before, this is still being discussed
+[here](https://github.com/WICG/display-locking/issues/150) and is subject to
+change.
+
+#### ScrollToTextFragment
+1. **Removed from DOM**: ScrollToTextFragment will not scroll to the removed
+element. If there is a second match which is still in the DOM after
+beforematch, it will be scrolled to but a second beforematch event will not be
+fired.
+2. **Reparented in DOM**: ScrollToTextFragment will scroll to the new
+location of the target. If there is a second match which was not reparented,
+ScrollToTextFragment will scroll to whichever comes first from the top of the
+page after the beforematch event is fired.
+3. **display: none**: ScrollToTextFragment will not scroll to the target
+element. If there is a second match which was not modified, ScrollToTextFragment
+will scroll to it.
+4. **visibility: hidden**: ScrollToTextFragment will not scroll to the target
+element. If there is a second match which was not modified, ScrollToTextFragment
+will scroll to it.
+5. **subtree-visibility: hidden-matchable -> visible**: ScrollToTextFragment
+will scroll to the unlocked text.
+
+#### ElementFragment
+This is the behavior both when navigating to a page with an element fragment in
+the url and when assigning to `window.location.hash`.
+1. **Removed from DOM**: The page will not scroll.
+2. **Reparented in DOM**: The page will scroll to the target element's new
+location.
+3. **display: none**: The page will not scroll.
+4. **visibility: hidden**: The page will scroll to the target element. This
+behavior makes sense because it matches the behavior when targeting a
+`visibility: hidden` element with a fragment regardless of beforematch.
+5. **subtree-visibility: hidden-matchable -> visible**: The page will scroll
+to the newly painted element.
+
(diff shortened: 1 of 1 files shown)


explainer-beforematch.md
```diff
@@ -191,6 +191,86 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
...
+`display: none` to the target element.
+4. **visibility: hidden**: The `beforematch` event handler adds the style
+`visibility: hidden` to the target element.
+5. **subtree-visibility: hidden-matchable -> visible**: The target element has
+the style `subtree-visibility: hidden-matchable` before the `beforematch` event
+is fired on the target element, and the `beforematch` event handler changes the
+style value from `hidden-matchable` to `visible`.
+
+
+Here is the current scrolling behavior in chromium for each of those cases in
```
vmpstr, line 217:
```suggestion
Here is the current scrolling behavior in Chromium for each of those cases in
```

explainer-beforematch.md
```diff
@@ -191,6 +191,86 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
...
+ScrollToTextFragment, and ElementFragment, and find-in-page:
+
+#### ScrollToTextFragment
+1. **Removed from dom**: ScrollToTextFragment will not scroll to the removed
+element. If there is a second match which is still in the dom after
+beforematch, it will be scrolled to.
```
vmpstr, line 223: Just to clarify, it will be scrolled to immediately, or will the algorithm for the search continue, encounter a new match, and then fire a new beforematch event?

explainer-beforematch.md
```diff
@@ -191,6 +191,86 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
...
+2. **Reparented in dom**: ScrollToTextFragment will scroll to the new
+location of the target. If there is a second match which was not reparented,
+ScrollToTextFragment will scroll to whichever comes first after the beforematch
```
vmpstr, line 226: Comes first in what order? From the top of the page? Or from where the initial match was?

explainer-beforematch.md
```diff
@@ -191,6 +191,86 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
...
+element. If there is a second match which was not modified, ScrollToTextFragment
+will scroll to it.
+4. **visibility: hidden**: ScrollToTextFragment will not scroll to the target
+element. If there is a second match which was not modified, ScrollToTextFragment
+will scroll to it.
+5. **subtree-visibility: hidden-matchable -> visible**: ScrollToTextFragment
+will scroll to the unlocked text.
+
+#### ElementFragment
+This is the behavior when navigating to a page with an element fragment, rather
```
vmpstr, line 238: Is there a separate behavior for fragment set by script?

explainer-beforematch.md
```diff
@@ -191,6 +191,86 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
...
+than the fragment being set by script.
+1. **Removed from dom**: The page will not scroll.
+2. **Reparented in dom**: The page will scroll to the target element's new
+location.
+3. **display: none**: The page will not scroll.
+4. **visibility: hidden**: The page will scroll to the target element. This
+behavior makes sense because it matches the behavior when targeting a
+`visibility: hidden` element with a fragment regardless of beforematch.
+5. **subtree-visibility: hidden-matchable -> visible**: The page will scroll
+to the activated element.
```
vmpstr, line 248:
```suggestion
to the newly painted element.
```

explainer-beforematch.md
```diff
@@ -191,6 +191,86 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
...
+
+#### find-in-page
+The find-in-page behavior has some issues in these cases. These will likely
+be fixed after adding an async step to pick up layout changes made by the
+beforematch event listener.
+1. **Removed from dom**: find-in-page does not scroll to the target element. If
+there is a second match in the DOM, find-in-page will immediately find it and
```
vmpstr, line 255: immediately here means without another beforematch?

explainer-beforematch.md
```diff
@@ -191,6 +191,86 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
...
+scroll to it.
+2. **Reparented in dom**: If the only match in the document is reparented to the
+end of the document, find-in-page becomes stuck at "0/x" matches and can't
+scroll to the new location, even if you keep typing more of the match out or
+close and reopen find-in-page. This behavior is obviously not good, and
```
vmpstr, line 260: > This behavior is obviously not good

Yep, we should probably fix it. 

There's references to an 'async step' here. Could you file a crbug and reference it here, just so that we can discuss and track progress?

explainer-beforematch.md
josepharhar: let me check...

josepharhar: It looks like there is no beforematch event fired on the second match. I figured it would be the other way around, but I'll just document it here and move on.

explainer-beforematch.md
josepharhar: From the top of the page. I'll clarify here.

explainer-beforematch.md
josepharhar: immediately means without having to press enter or anything to go to the next match. I'll clarify here

explainer-beforematch.md
josepharhar: I made a bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1074121
and linked it at the text under ####find-in-page

explainer-beforematch.md
josepharhar: Good question... I'll have to test it out.

josepharhar: It's the same for both, I clarified here.

explainer-beforematch.md
```diff
@@ -191,6 +191,88 @@ In general, we believe that the `beforematch` event does not pose a privacy prob
...
+`display: none` to the target element.
+4. **visibility: hidden**: The `beforematch` event handler adds the style
+`visibility: hidden` to the target element.
+5. **subtree-visibility: hidden-matchable -> visible**: The target element has
+the style `subtree-visibility: hidden-matchable` before the `beforematch` event
+is fired on the target element, and the `beforematch` event handler changes the
+style value from `hidden-matchable` to `visible`.
+
+
+Here is the current scrolling behavior in Chromium for each of those cases in
```
vmpstr, line 217: I'd add a small disclaimer here, something like

This is subject to change as we evolve and improve the implementation.

josepharhar: done.

josepharhar:
> Thanks! Also feel free to file crbugs for behaviors that you think should be changed, and issues for behaviors that you think warrant some discussion before we make code changes.
> 
> As an example, not firing a second beforematch if the first patch is removed from the dom: I'm honestly not sure what the right thing here is. Should we fire the second event? I could be convinced either way, and will have to think about it for a bit...

I figured this would be addressed here: https://github.com/WICG/display-locking/issues/150

PR #155, opened 2020-04-24
rename to content-visibility

vmpstr:
This renames the feature from subtree-visibility to content-visibility. The term subtree is still used and defined here, which I think is useful, since just talking about content may be ambiguous.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,6 +1,6 @@
 <pre class='metadata'>
-Title: CSS Subtree Visibility
-Shortname: css-subtree-visibility
+Title: CSS Content Visibility
+Shortname: css-content-visibility
 Level: 1
 Status: UD
 Group: csswg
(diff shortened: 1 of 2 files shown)


vmpstr:
Please take a look.

This is just a straight-up rename. I think it still reads the same without further changes to terms like 'subtree', but let me know if I should change those to something as well.

yuanyan:
`content-visibility` seems like `visibility` 😄

kenchris:
You should fix the readme as well

image

vmpstr:
Oh, I completely forgot about the readme, will do thanks!

Re naming: it does serve a similar purpose to `visibility` :), but also has some differences

vmpstr:
The rest of the renames is in https://github.com/WICG/display-locking/pull/156

Issue #160, opened 2020-05-02
beforematch: Target element definition is ambiguous.


emilio:
It says:

> the element that contains the text which was matched by find-in-page or scroll-to-text navigation.

What happens when the text to be found spans multiple elements, like:

`<span>foo</span> <span>bar</span>`

What if the match happens across shadow tree boundaries?

josepharhar:
I agree this is ambiguous, I need to update the explainer.

The event is fired on the block-level element containing the text which will be scrolled to. In the case of `<div id=parent> <span>foo</span> <span>bar</span> </div>`, beforematch will be fired on `parent` in the scroll-to-text and find-in-page cases.

As for shadow boundaries, the beforematch event is fired on the first block-level element containing the matching text in the flat tree. I just tested it out and it appears to basically ignore shadow boundaries, the event is fired on whatever the first block-level containing element is, whether its in shadow or not. If the shadow boundary is inline and the matching text goes across it, it just goes to the first block-level containing element which in this case would be outside of the shadow dom.

josepharhar:
@emilio I updated the explainer to elaborate some more. Has the ambiguity been resolved?

emilio:
What does "block-level ancestor element" mean in this case? An inline-block is a block inside, but not outside, for example. Also probably should reference "block level boxes".

The new definition is a bit concerning, consider you have an inline element with a shadow root, with a bunch of text in it. The closest block ancestor is outside the shadow root, but it doesn't know it has any text on it. Seems like this breaks shadow boundary encapsulation by dispatching events outside of a shadow tree for stuff that happens inside of the shadow tree.

josepharhar:
Thanks for the feedback!

> What does "block-level ancestor element" mean in this case? An inline-block is a block inside, but not outside, for example. Also probably should reference "block level boxes".

I tested this out with this HTML:
```html
<div id=one style="display: block">
  one 
  <div id=two style="display: inline-block">
    two 
    <div id=three style="display: inline">
      three
    </div>
  </div>
</div>
```
I found that whatever text is searched for, `<div id=one>` is always selected to fire beforematch on. Is this the case you are talking about, or is it something else? Is `<div id=two>` a block level box?

[...]

PR #164, opened 2020-05-05
Elaborate summary in beforematch explainer

josepharhar:
Addresses https://github.com/WICG/display-locking/issues/160

### DIFF
--- a/explainer-beforematch.md
+++ b/explainer-beforematch.md
@@ -3,21 +3,28 @@
 ### Summary (TL;DR)
 
 The `beforematch` event allows developers to display content to the user in
-response to find-in-page or fragment navigation actions. It is fired just
-before these actions, and at render timing. The applicable actions are the
-following:
+response to actions which scroll the page to a target element. It is fired
+at render timing just before these actions scroll the page. The applicable
+actions are the following:
 
-* There is a new find-in-page active match, in which case the event is fired
-  with the matched element as the target.
-* There is a fragment-link navigation, in which case the event is fired with
-  the matched fragment as the target.
-* There is a scroll-to-text navigation, in which case the event is fired with
+* There is a new find-in-page (ctrl+f) active match, in which case the event
+  is fired with the matched element as the target.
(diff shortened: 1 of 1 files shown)


explainer-beforematch.md
```diff
@@ -3,21 +3,25 @@
...
+  is fired with the matched element as the target.
+* There is a [scroll-to-text](https://github.com/WICG/ScrollToTextFragment)
+  navigation (`example.com/#:~:text=foo`), in which case the event is fired with
   the matched element as the target.
 
 Note that the 'matched element' in this document refers to one of the following:
-* the element that contains the text which was matched by find-in-page or
-  scroll-to-text navigation.
-* the element which was selected for the fragment link navigation.
+* In the find-in-page and scroll-to-text cases: The closest block-level element
```
vmpstr, line 20:
```suggestion
* In the find-in-page and scroll-to-text cases: The closest block-level ancestor element
```

Also, I would specify here that if the matched text spans multiple block-level elements, then the event is fired on the block-level ancestor of the beginning of the match
(possibly worded better)

josepharhar: matching text spanning multiple block level elements is only possible with scroll-to-text-fragment right? in that case, didn't we earlier discuss making it only fire on the first of the two block level elements? I guess we could change it to fire on the ancestor...

vmpstr: You're right, I meant to say that there should be two changes here:
- Specify that it's the closest block-level **ancestor** element (_ie clarify what "closest" means by saying that we're talking about ancestor distance here_)
- Specify that if it spans multiple blocks, then use **the beginning of the match** as the spot from which you look for the ancestor element

Sorry if that wasn't clear. I am _not_ suggesting a **common** ancestor for the range of the match. Does that make sense?

josepharhar: got it, thanks!

josepharhar: How does the latest patch look?

PR #165, opened 2020-05-11
Update links to point to csswg-drafts

vmpstr:


vmpstr:
Please take a look.

chrishtr:
What about deleting the specs here and replacing them with a redirect link to the CSSWG version?

For both contain-intrinsic-size and content-visibility.

vmpstr:
Sounds good. Done. (expand the .html diffs to see the redirect)

PR #168, opened 2020-05-28
Restructured the documentation

vmpstr:

PR #169, opened 2020-05-28
Rename explainer files to align with links

vmpstr:


vmpstr:
This is a follow-up to #168 where I forgot to rename two files. The links have changed but the files haven't been renamed; oops. This one should fix everything up.

PR #173, opened 2020-06-04
Add alternatives section to beforematch explainer

josepharhar:
I made example one more complicated in order to make some points i the alternatives, I hope thats OK

### DIFF
--- a/explainers/beforematch.md
+++ b/explainers/beforematch.md
@@ -67,20 +67,39 @@ observations.
 .collapsed {
   content-visibility: hidden-matchable;
 }
+.title-collapsed::before {
+  content: '➡';
+}
+.title-open::before
+  content: '⬇';
+}
 </style>
 
 Please explore the following sections:
-<h1>Introduction</h1> <div class=collapsed>lorem ipsum ...</div>
-<h1>Thesis</h1>       <div class=collapsed>dolor sit amet ...<div>
-<h1>Conclusion</h1>   <div class=collapsed>consectetur adipiscing ...</div>
+<div class=section>
+  <h1 class="title-collapsed">Introduction</h1>
+  <div class=collapsed>lorem ipsum ...</div>
+</div>
+
+<div class=section>
+  <h1 class="title-collapsed">Thesis</h1>
+  <div class=collapsed>dolor sit amet ...<div>
+</div>
+
+<div class=section>
+  <h1 class="title-collapsed">Conclusion</h1>
+  <div class=collapsed>consectetur adipiscing ...</div>
+</div>
 
 <script>
-function expand(e) {
-  e.target.classList.remove("collapsed");
-}
-
-document.querySelectorAll(".collapsed").forEach(item => {
-  item.addEventListener("beforematch", expand);
+document.querySelectorAll('.section').forEach(section => {
+  const title = section.querySelector('.title-collapsed');
+  const hiddenContent = section.querySelector('.collapsed');
+  section.addEventListener('beforematch', () => {
+    hiddenContent.classList.remove('collapsed');
+    title.classList.remove('title-collapsed');
+    title.classList.add('title-open');
+  });
 });
 </script>
 ```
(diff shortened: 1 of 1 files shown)


explainers/beforematch.md
```diff
@@ -288,6 +307,50 @@ scroll after adding an async step.
...
+
+#### Automatic Unlocking with Activation Event
+This is like "Automatic Unlocking," but with an added "activation" emitted when
+content is displayed to allow developers to change other state and styles if
+needed.
+##### Pros
+* Same pros as "Automatic Unlocking."
+* Allows other state/styles to be changed when content is displayed.
+##### Cons
+* TODO: ask vmpstr and chrishtr for more context on this
```
josepharhar, line 335: What was the reason not to go with activation events again?

vmpstr: So it depends how you look at it
- On the one hand, activation event goes hand-in-hand with automatic unlocking/expanding, which we think is not a good use-case since the majority of the time the developer would want to do something other than just show the element. That, and it makes state tracking hard (ie the developer doesn't have the tools to collapse the section again; we would have to add something like a JS api). In this case activation event informs that this has happened. However, since we're not doing the automatic unlocking then activation event is not needed.

- On the other hand, activation event could be fired prior to us finding a match that would have otherwise unlocked the element. That's exactly what beforematch is :) In other words, we renamed activation event to beforematch when we decided not to automatically unlock due to reasoning above.

Does that make sense?

explainers/beforematch.md
```diff
@@ -288,6 +307,50 @@ scroll after adding an async step.
 5. **content-visibility: hidden-matchable -> visible**: find-in-page scrolls to
 to the revealed text.
 
+### Alternatives
+Given the purpose of displaying `content-visibility: hidden-matchable` text
+when it is searched for, there are a number of alternatives we have considered.
+
+#### Automatic Unlocking
+`content-visibility: hidden-matchable` text would be automatically changed to
+`content-visibility: visible` when the hidden text is searched for.
```
vmpstr, line 316: This is not actually correct. We would keep an internal state that essentially ignores 'hidden-matchable'. The problem is that, unlike attributes, we can't change styles from within the browser. 

If we have something like
```
body > .randomClass {
  content-visibility: hidden-matchable;
}

it means that all first level children of body that have .randomClass have hidden-matchable. If we find one such element, we obviously can't change this style rule since it doesn't just apply to this element. We could change inline style of the element, but that sounds like a really bad approach since we don't know how script is using inline styles, so we shouldn't be messing with them.

File: explainers/beforematch.md [hunk 3]
```diff
@@ -288,6 +307,50 @@ scroll after adding an async step.
...
+##### Pros
+* Provides the desired behavior without the need for the web developer to do
+  anything besides use the `hidden-matchable` value.
+* Less complex than `beforematch` with regards to scrolling behavior.
```
vmpstr, line 320: We need to elaborate on this: Specifically, since script does not run and the interaction is fully within browser control, we can guarantee that scrolling to the element is possible

explainers/beforematch.md
```diff
@@ -288,6 +307,50 @@ scroll after adding an async step.
...
+##### Cons
+* Doesn't allow the developer to change other properties in conjunction with
```
vmpstr, line 322: One other con is what I said before: we need to have some internal state tracking that isn't obvious

explainers/beforematch.md
vmpstr: Just to add to this as I'm reading the doc, the important con is that automatically changing CSS isn't possible, so there is a lot of complexity involved with automatic unlocking. And the whole point of this activation event here is that if we're firing an activation event anyway, then why bother with automatic unlocking? Just let the script take care of it and reduce the complexity and internal states in the browser.

To put it differently, automatic unlocking with an event is not any better than just the event, because the handler in the event could still remove the element, relock the element, move it, etc, etc. All of those need to be dealt with whether or not we automatically unlock in the presence of an event.

explainers/beforematch.md
```diff
@@ -288,6 +307,50 @@ scroll after adding an async step.
...
+* Same pros as "Automatic Unlocking."
+* Allows other state/styles to be changed when content is displayed.
+##### Cons
+* TODO: ask vmpstr and chrishtr for more context on this
+
+#### CSS Pseudo Selector
+A pseudo selector, such as `:target`, would be applied to the element
+containing the matching text when it is searched for.
+##### Pros
+* Allows content to become visible when searched for with only CSS.
```
vmpstr, line 341: Also allows custom styling on the element

explainers/beforematch.md
```diff
@@ -288,6 +307,50 @@ scroll after adding an async step.
...
+needed.
+##### Pros
+* Same pros as "Automatic Unlocking."
+* Allows other state/styles to be changed when content is displayed.
+##### Cons
+* TODO: ask vmpstr and chrishtr for more context on this
+
+#### CSS Pseudo Selector
+A pseudo selector, such as `:target`, would be applied to the element
+containing the matching text when it is searched for.
```
vmpstr, line 339: Also possibly the same or different selector on the ancestor chain to allow for different styling

explainers/beforematch.md
```diff
@@ -288,6 +307,50 @@ scroll after adding an async step.
...
+* Allows content to become visible when searched for with only CSS.
+##### Cons
+* If CSS with a pseudo selector is used to make text visible, then when
+  find-in-page is closed or the search text changes, the pseudo selector would
+  be removed and then any selector which is displaying the text based on that
+  pseudo selector would not apply, causing the expanded section to unexpectedly
+  collapse.
+* There is no way in CSS to say "change my style if a descendant has a pseudo
+  class on it." For this reason, developers would be unable to change styles
+  outside of the particular element that has the matching text, which would
```
vmpstr, line 350: This isn't necessarily true if we add the pseudo class on the ancestor hierarchy. I think the point here is that it can be complicated or impossible to provide the right selector which can modify a style on an unrelated element (Such as the case in the title-collapsed or title-expanded that you have above

explainers/beforematch.md
```diff
@@ -288,6 +307,50 @@ scroll after adding an async step.
...
+##### Cons
+* TODO: ask vmpstr and chrishtr for more context on this
+
+#### CSS Pseudo Selector
+A pseudo selector, such as `:target`, would be applied to the element
+containing the matching text when it is searched for.
+##### Pros
+* Allows content to become visible when searched for with only CSS.
+##### Cons
+* If CSS with a pseudo selector is used to make text visible, then when
```
vmpstr, line 343: Is it possible for you to take a few screenshots on a phone interaction with mobile wiki page
- One screen of collapsed sections
- One screen of find-in-page that finds text in a collapsed section (you can manually expand the section for the purpose of the demo)
- One screen with find-in-page dialog dismissed that still displays the open section.

I think that example would make this con point obvious

vmpstr: Nevermind, I see you added a similar thing in the #172

explainers/beforematch.md
josepharhar: Why isn't automatically changing CSS possible? If we are able to fire an event that has the power to change CSS, then can't we just change the style from `hidden-matchable` to `visible` in C++?

explainers/beforematch.md
josepharhar: I reworded this bullet point. Look good?

explainers/beforematch.md
josepharhar: Done

explainers/beforematch.md
josepharhar: done

explainers/beforematch.md
vmpstr: Which style do we change? We can't change the rule style because the rule applies to more than one element (in the general case; ie * { content-visibility: hidden-matchable } applies to _all_ elements, so when we activate, we can't just change that to visible since that un-hides everything). 

The only possibility is that we set the inline style of the element to have content-visibility: visible. But again, if for example the descendant elements have content-visibility: inherit or something then we're changing more than just the one element. That, and changing the inline style automatically open the door to a lot of web breaking, since if script generates and sets a new inline style, all of a sudden it overrides the values we set. That's just not going to be a good change.

The idea of the event is to defer this decision to script, since there is no one correct decision in terms of what to do, so we have to delegate the decision to the app.

explainers/beforematch.md
vmpstr: Yep, this looks good.

explainers/beforematch.md
josepharhar: Ah I see, having the browser change styles doesn't really make sense. Thanks for the explanation.
I added another alternative to show why this is a bad idea.

explainers/beforematch.md
josepharhar: Thanks for the explanation, I resolved this in another thread

explainers/beforematch.md
josepharhar: Done

explainers/beforematch.md
josepharhar: Thanks for the explanation, I put your words into this bullet point

PR #174, opened 2020-06-05
Find in page compat

vmpstr:
Document the find-in-page functionality in various scenarios across Chrome, Firefox, and Safari.


vmpstr:
Please take a look.

PR #176, opened 2020-07-26
Fix markdown syntax

azu:
add missing backtick.


vmpstr:
Marked as non substantive for IPR from ash-nazg.

PR #180, opened 2020-08-14
Beforematch spec draft

vmpstr:
This adds an initial draft of beforematch and hidden-matchable spec

### DIFF
--- a/explainers/beforematch-spec-draft.md
+++ b/explainers/beforematch-spec-draft.md
@@ -1,0 +1,51 @@
+# beforematch event
+
+_These changes will likely go into the
+[find-in-page](https://html.spec.whatwg.org/multipage/interaction.html#find-in-page)
+section of the HTML spec_
+
+When a new active match is set, either by advancing through the match list or
+due to a new find-in-page request, a `beforematch` event is fired on a node
+identified by the active match [range
+start](https://dom.spec.whatwg.org/#concept-range-start). This process follows
+the following algorithm:
+
+1. Identify the match _M_ which will become the new active match. _M_'s DOM
+range must not be [collapsed](https://dom.spec.whatwg.org/#range-collapsed).
+
+2. At the next rendering opportunity, specifically step 12 of [Update the
+rendering](https://html.spec.whatwg.org/#rendering-opportunity) step:
+
+  2.1. If the range that constitutes match _M_ is
+      [collapsed](https://dom.spec.whatwg.org/#range-collapsed), restart the
+      algorithm at Step 1.
+
+  2.2. Fire a `beforematch` event on an element identified by _M_'s range start node.
+
+3. If the event fired in Step 2.2 was handled by any beforematch handler, then
+delay Step 4 until the next rendering opportunity (the next Step 12 of [Update the
+rendering](https://html.spec.whatwg.org/#rendering-opportunity)). Otherwise,
+proceed to Step 4 immediately.
+
+4. If the range that consitutes match _M_ is
+[collapsed](https://dom.spec.whatwg.org/#range-collapsed), or the range start
+node is not shown to the user (e.g. it is inside a `content-visibility:
+hidden-matchable` subtree), restart the algorithm at Step 1 skipping match _M_.
+
+5. Scroll the range start node of match _M_ into view.
+
+# content-visibility: hidden-matchable
+
+_These changes will likely go into the
+[css-contain-2](https://www.w3.org/TR/css-contain-2/#content-visibility)
+module_.
+
+'hidden-matchable': 
+  The element skips its contents.
+
+  The skipped contents must not be accessible to user-agent features such as
+  tab-order navigation, nor be selectable or focusable.
+
+  The skipped contents must be accessible to the find-in-page algorithm in order
+  to allow the `beforematch` event to fire. (TODO: reference the find-in-page
+  beforematch spec)
--- a/explainers/beforematch-spec-draft.md
+++ b/explainers/beforematch-spec-draft.md
@@ -10,7 +10,7 @@ identified by the active match [range
 start](https://dom.spec.whatwg.org/#concept-range-start). This process follows
 the following algorithm:
 
-1. Identify the match _M_ which will become the new active match. _M_'s DOM
+1. Identify the candidate match _M_ which will become the new active match. _M_'s DOM
 range must not be [collapsed](https://dom.spec.whatwg.org/#range-collapsed).
 
 2. At the next rendering opportunity, specifically step 12 of [Update the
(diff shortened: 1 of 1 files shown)


vmpstr:
Please take a look.

explainers/beforematch-spec-draft.md
```diff
@@ -0,0 +1,51 @@
...
+section of the HTML spec_
+
+When a new active match is set, either by advancing through the match list or
+due to a new find-in-page request, a `beforematch` event is fired on a node
+identified by the active match [range
+start](https://dom.spec.whatwg.org/#concept-range-start). This process follows
+the following algorithm:
+
+1. Identify the match _M_ which will become the new active match. _M_'s DOM
+range must not be [collapsed](https://dom.spec.whatwg.org/#range-collapsed).
```
josepharhar, line 14: There isn't any code that looks at whether the range is collapsed. If the matching range is collapsed at the time of doing the search, then we simply wouldn't have found it, because its empty...?
Or is this not the same time as when the search is performed?
I mean I guess it makes sense to point this out, I just wonder if maybe it could be more clear that there is no need to actually check collapsed at this point? I guess saying "must not be" sort of implies this...?

explainers/beforematch-spec-draft.md
```diff
@@ -0,0 +1,51 @@
...
+
+2. At the next rendering opportunity, specifically step 12 of [Update the
+rendering](https://html.spec.whatwg.org/#rendering-opportunity) step:
+
+      2.1. If the range that constitutes match _M_ is
+        [collapsed](https://dom.spec.whatwg.org/#range-collapsed), restart the
+        algorithm at Step 1.
```
josepharhar, line 21: We also restart the search next time starting after the range which got collapsed. The collapsed range is passed in to the next search the same way that selection is passed into find-in-page - when you select some text on a page and then start a find-in-page, it starts searching from (the beginning?) of the text the user selected.

If this is hard to define then maybe we could leave it out for now.

explainers/beforematch-spec-draft.md
```diff
@@ -0,0 +1,51 @@
...
+
+      2.2. Fire a `beforematch` event on an element identified by _M_'s range start node.
+
+3. If the event fired in Step 2.2 was handled by any beforematch handler, then
+delay Step 4 until the next rendering opportunity (the next Step 12 of [Update the
+rendering](https://html.spec.whatwg.org/#rendering-opportunity)). Otherwise,
+proceed to Step 4 immediately.
```
josepharhar, line 28: I haven't implemented this step or planned on doing so.

I suppose I could give it a shot, but it would require carefully traversing the event bubbling route that the event will take up the DOM tree and checking for beforematch handlers all the way.

Do you think this is important to do? I suppose it would make find-in-page one frame faster in most normal cases which don't involve beforematch, but I *personally* haven't noticed any slowness from the double rAF I put in m86.

explainers/beforematch-spec-draft.md
```diff
@@ -0,0 +1,51 @@
...
+
+4. If the range that consitutes match _M_ is
+[collapsed](https://dom.spec.whatwg.org/#range-collapsed), or the range start
+node is not shown to the user (e.g. it is inside a `content-visibility:
+hidden-matchable` subtree), restart the algorithm at Step 1 skipping match _M_.
+
+5. Scroll the range start node of match _M_ into view.
```
josepharhar, line 35: To do the scrolling, TextFinder uses a LayoutObject scrolling method which is along the code path for element.scrollIntoView(), so this is basically the same as calling element.scrollIntoView(). Not sure if it's worth saying explicitly here, but I suppose you could if you wanted.

explainers/beforematch-spec-draft.md
```diff
@@ -0,0 +1,51 @@
...
+
+# content-visibility: hidden-matchable
+
+_These changes will likely go into the
+[css-contain-2](https://www.w3.org/TR/css-contain-2/#content-visibility)
+module_.
+
+'hidden-matchable': 
+  The element skips its contents.
```
josepharhar, line 44: This sentence seems like its lacking context on its own, but I'm guessing it makes sense in the context of the referenced css-contain-2

explainers/beforematch-spec-draft.md
chrishtr: Do you mean the "otherwise" part? I think it's ok to unconditionally delay.

explainers/beforematch-spec-draft.md
chrishtr: Add link to "skip" definition

chrishtr: Perhaps rephrase to "The element always skips its contents in an equivalent way to the 'hidden' value"

explainers/beforematch-spec-draft.md
josepharhar: Yes, we are unconditionally delaying right now, and it would be a little complicated to check if we actually called a beforematch handler due to event bubbling.

explainers/beforematch-spec-draft.md
vmpstr: Yeah this is already naturally the case. This is an important point only because of the step that restarts at step 1 again when we're collapsed. Other implementations may consider collapsed sections as viable matches? I don't really know, but if that's the case then we'd get into an infinite loop, hence me saying that must not be collapsed at this point.

josepharhar: lgtm

explainers/beforematch-spec-draft.md
vmpstr: Ok, I think I see what you're getting at. I'll add a note here.

explainers/beforematch-spec-draft.md
vmpstr: Sounds good, I will remove the condition.

explainers/beforematch-spec-draft.md
vmpstr: I can point to it for sure!

explainers/beforematch-spec-draft.md
vmpstr: Sounds good!

explainers/beforematch-spec-draft.md
josepharhar: One of my patched in review right now also makes it so that during this circular finding process stops if it goes past the first range we matched by storing the first range and passing it through the AsyncScrollContext and the Find function itself. Also not sure if this is worth mentioning now or if it would be hard to define, but fyi.

vmpstr:
PTAL

PR #184, opened 2020-09-25
Combine hidden-matchable and beforematch explainers

vmpstr:


vmpstr:
Please take a look. This is sourced mostly from the beforematch explainer, with a couple of notes about hidden-matchable.

josepharhar (approved):
lgtm!
I think I definitely need to revise some of the things I've said, but maybe I'll do that after privacy discussions which could make another revision needed

vmpstr:
> I think I definitely need to revise some of the things I've said, but maybe I'll do that after privacy discussions which could make another revision needed

Agreed. I tried to leave it as "it is now" but left a few sentences about the fact that we are still evolving the API with considerations from both TAG and privacy

PR #185, opened 2020-10-02
Update beforematch explainer

josepharhar:
I did several things in this patch:
- Made beforematch.md refer to hidden-content-explainer.md.
- Removed element fragment navigation beforematch feature.
- Rewrote Privacy Concerns section


explainers/hidden-content-explainer.md
```diff
@@ -12,28 +12,21 @@ This is an explainer for two closely related features:
...
-* In the find-in-page and scroll-to-text cases: The closest block-level ancestor
-  element in the flat tree containing the matching text.
-  If the match spans multiple block-level elements, then the first block-level
-  element is used.
-  Since the flat tree is used to determine the element to fire the event on,
-  shadow boundaries have no impact.
-* In the element fragment case: The element which was selected by id from the
-  fragment.
+Note that the 'matched element' in this document refers to the closest
+block-level ancestor element in the flat tree containing the matching text.
```
vmpstr, line 26: Instead of matching text, say "match" and link to https://html.spec.whatwg.org/multipage/interaction.html#fip-matches

explainers/hidden-content-explainer.md
```diff
@@ -135,56 +128,45 @@ sections already has an event handler to expand and show the section. With the
...
-
-Some consideration needs to be given to the `beforematch` event when it comes to
-cross frame interactions. Specifically, if the matched information is found within
-a frame, should the signal propagate to the embedding page?
-
-In general, we believe that the `beforematch` event does not pose a privacy problem.
-However, the privacy aspect of the event should be discussed in more detail via
-formal security reviews.
+The beforematch event could expose more information to the page than is
+currently exposed because it very directly hints to the text the user is
```
vmpstr, line 132: I would simplify this part to say something like

The beforematch event exposes more information to the page than is currently exposed because it is fired on an element that contains the text the user is searching for.

explainers/hidden-content-explainer.md
```diff
@@ -135,56 +128,45 @@ sections already has an event handler to expand and show the section. With the
...
+searching for since it gets fired on the element containing the target text.
+
+Note that it is already possible to snoop on find-in-page by creating a very
+small scrollable area containing every next character the user could type into
```
vmpstr, line 136: s/a very small/0 height/ 

Or I would just say this is separate movements: 
- It is possible to snoop this by creating a scrollable area containing every next ...
- Furthermore, it is possible to make this area very small (0px or 1px tall) so that the searched text is not visible to the user

explainers/hidden-content-explainer.md
```diff
@@ -135,56 +128,45 @@ sections already has an event handler to expand and show the section. With the
...
+find-in-page, listening to scroll events to see which character the user typed
+in, then prepending the new character to all of the next possible search terms.
+This is demonstrated in
+[search-incremental.html](/resources/find-in-page/search-incremental.html).
```
vmpstr, line 140: Is this the right URL for this? Don't you need to have this as a relative path since it will be something like:
/WICG/display-locking/tree/master/resources/find-in-page

explainers/hidden-content-explainer.md
```diff
@@ -135,56 +128,45 @@ sections already has an event handler to expand and show the section. With the
...
+
+In order to mitigate the additional information that beforematch could expose to
+the page, we have added two constraints to when beforematch can be fired:
```
vmpstr, line 143: .. two constraints to beforematch

explainers/hidden-content-explainer.md
```diff
@@ -135,56 +128,45 @@ sections already has an event handler to expand and show the section. With the
...
+
+1. The beforematch event will only be fired on `content-visibility:
```
vmpstr, line 145: on content in the DOM subtree of an element with content-visibility: hidden-matchable.

(it's a bit unclear what content-visibility: hidden-matchable content is)

explainers/hidden-content-explainer.md
```diff
@@ -135,56 +128,45 @@ sections already has an event handler to expand and show the section. With the
...
+hidden-matchable` content. This removes the information delta
+of find-in-page on normal pages without hidden find-in-page snooping. Without
+constraint, a page could listen to beforematch events and get some
+idea of what you’re searching for within the text already in the page. Although
+a page could already try to do this by listening to scroll events and guessing
+which element was scrolled to, the beforematch event is a bit of a stronger and
```
vmpstr, line 151: the beforematch event, without this restriction, is a stronger and more accurate signal.

explainers/hidden-content-explainer.md
```diff
@@ -135,56 +128,45 @@ sections already has an event handler to expand and show the section. With the
...
+more accurate signal in this case. By firing beforematch only on
+hidden-matchable content, we eliminate this possibility.
+
+2. If the page doesn’t reveal hidden-matchable content in response to a
```
vmpstr, line 155: We should say here what we mean by "reveal". We are only referring to the fact that after beforematch event handler runs, the text is no longer under a content-visibility: hidden-matchable, or display: none subtrees and it has visibility: visible style. We don't attempt to detect other ways of hiding content such as overflow clipping, opacity or color manipulation

explainers/hidden-content-explainer.md
```diff
@@ -135,56 +128,45 @@ sections already has an event handler to expand and show the section. With the
...
+beforematch event, the beforematch event will not be fired for the remainder of
+the lifetime of the document. This will prevent the page from building out a
+string of what the user is searching for without having to reveal that string
+to the user visually. Although this is already possible as shown in
+[search-incremental.html](/resources/find-in-page/search-incremental.html),
+firing beforematch on hidden content that stays hidden would be harder for the
+user to notice.
+
+As for the ScrollToTextFragment case, the page can already guess with reasonable
```
vmpstr, line 164: FWIW ScrollToTextFragment also just happens once on the page, so we also only fire the event in the hidden-matchable subtree. An interesting question is what happens if the page doesn't reveal it: does it mean that find-in-page also won't fire the event because of restriction 2? Or is scroll to text fragment's beforematch event logically different from find-in-page beforematch event?

explainers/hidden-content-explainer.md
```diff
@@ -12,28 +12,21 @@ This is an explainer for two closely related features:
...
 
-2. The `beforematch` event allows developers to display content to the user in
-    response to actions which scroll the page to a target element. It is fired
-    at render timing just before these actions scroll the page. The applicable
-    actions are the following:
+2. The `beforematch` event allows developers to display `hidden-matchable`
+    content to the user in response to searches which scroll the page to a
+    target element. It is fired at render timing before these actions scroll
+    the page. The applicable actions are the following:
     * There is a new find-in-page (ctrl+f) active match, in which case the event
```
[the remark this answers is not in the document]
josepharhar: done

explainers/hidden-content-explainer.md
```diff
@@ -12,28 +12,21 @@ This is an explainer for two closely related features:
...
       is fired with the matched element as the target.
-    * There is an element fragment navigation (`example.com/#foo`) or the fragment
-      is changed by script (`window.location.hash = ...`), in which case the event
-      is fired with the matched element as the target.
     * There is a [scroll-to-text](https://github.com/WICG/ScrollToTextFragment)
```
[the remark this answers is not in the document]
josepharhar: done

explainers/hidden-content-explainer.md
josepharhar: I removed this text and put a link to fip-matches in the find-in-page bullet point. how does it look?

explainers/hidden-content-explainer.md
josepharhar: done

explainers/hidden-content-explainer.md
josepharhar: I reorganized it sort of like you suggested, how does it look?

explainers/hidden-content-explainer.md
josepharhar: This appears to work: https://github.com/josepharhar/display-locking/blob/update/explainers/hidden-content-explainer.md

explainers/hidden-content-explainer.md
josepharhar: done

explainers/hidden-content-explainer.md
josepharhar: done

explainers/hidden-content-explainer.md
josepharhar: done

explainers/hidden-content-explainer.md
josepharhar: I elaborated on "reveal." How does it look?

explainers/hidden-content-explainer.md
josepharhar: I clarified this, how does it look?

explainers/hidden-content-explainer.md
```diff
@@ -12,28 +12,21 @@ This is an explainer for two closely related features:
...
-    * There is a new find-in-page (ctrl+f) active match, in which case the event
-      is fired with the matched element as the target.
-    * There is an element fragment navigation (`example.com/#foo`) or the fragment
-      is changed by script (`window.location.hash = ...`), in which case the event
-      is fired with the matched element as the target.
+2. The `beforematch` event allows developers to display `hidden-matchable`
+    content to the user in response to searches which scroll the page to some
+    target text. The event is fired at render timing on the nearest ancestor
+    block-level element for these cases:
+    * There is a new find-in-page (ctrl+f)
```
vmpstr, line 19: You don't really need "ctrl+f" especially since it's "cmd+f" on mac :) I think find-in-page is unambiguous enough

explainers/hidden-content-explainer.md
```diff
@@ -12,28 +12,21 @@ This is an explainer for two closely related features:
...
-  If the match spans multiple block-level elements, then the first block-level
-  element is used.
-  Since the flat tree is used to determine the element to fire the event on,
-  shadow boundaries have no impact.
-* In the element fragment case: The element which was selected by id from the
-  fragment.
+      navigation (`example.com/#:~:text=foo`), where the target text is located
+      inside a `content-visibility: hidden-matchable` element.
+
+If the matching text spans multiple block-level elements, then the first
```
vmpstr, line 27: We need to decide whether we fire an event on the nearest ancestor or nearest ancestor with content-visibility: hidden-matchable (and make that clear)

explainers/hidden-content-explainer.md
vmpstr: TIL

explainers/hidden-content-explainer.md
vmpstr: Looks good

explainers/hidden-content-explainer.md

explainers/hidden-content-explainer.md
josepharhar: hmmm now that we are only firing on hidden-matchable, i suppose it could make sense to only fire on that element instead of going to the block-level element and bubbling it from there. what do you think?
Or are you just saying for the case where it spans multiple block level elements?

explainers/hidden-content-explainer.md
vmpstr: Looks good

explainers/hidden-content-explainer.md
vmpstr: Looks good.

explainers/hidden-content-explainer.md
vmpstr: 👍

explainers/hidden-content-explainer.md
vmpstr: Either way, I think we should find the nearest content-visibility: hidden-matchable element and fire it on that element (with bubbling). Same caveats apply: if the text spans multiple block level elements all of which have hidden-matchable, then just do it on the first one. I think it would be a rare case though, and I'm not sure we can find text across layout containment which hidden-matchable would put in.

josepharhar:
Does this look finished @vmpstr?

vmpstr (approved):

explainers/hidden-content-explainer.md
josepharhar: Yeah that's fair if we are only firing it on hidden-matchable content now.
I updated this and the previous text to say that we will only fire it on hidden-matchable content.

PR #188, opened 2020-11-05
Clarify responses to DOM changes in hidden content explainer

josepharhar:
Addresses #150

### DIFF
--- a/explainers/hidden-content-explainer.md
+++ b/explainers/hidden-content-explainer.md
@@ -172,11 +172,11 @@ the page, we have added two constraints to beforematch:
    will be disabled for the remaining lifetime of the document. By "reveal," I
    mean that the active match fulfills all of these requirements:
    - The DOM range of the target match is not
-     [collapsed](https://dom.spec.whatwg.org/#range-collapsed) (meaning that the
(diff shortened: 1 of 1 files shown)


explainers/hidden-content-explainer.md
```diff
@@ -167,11 +167,17 @@ the page, we have added two constraints to beforematch:
...
+2. If the page fails to reveal the active match when the browser tries to scroll
+   to the active match after firing the beforematch event, the beforematch event
+   will be disabled for the remaining lifetime of the document. By "reveal," I
+   mean that the active match fulfills all of these requirements:
+   - The DOM range of the target match is not
+     [collapsed](https://dom.spec.whatwg.org/#range-collapsed) (meaning that the
+     target match was not removed from the DOM.
+   - The `content-visibility` CSS property is not `hidden` or
+     `hidden-matchable`.
+   - The `display` CSS property is not `display: none`.
```
vmpstr, line 179: The visibility property is inherited, so the text for it is correct. For display none and content-visibility hidden, however, we should say that the element does not have display: none and is not in the subtree of any display: none element

PR #201, opened 2021-09-10
update rendering details

vmpstr:


vmpstr:
Please take a look.

explainers/update-rendering.md
```diff
@@ -42,12 +48,49 @@ are ready to be displayed to screen.  Note that repeated calls to
 updateRendering() returns new promises, with the ultimate priority of the
 request being the highest requested priority.
 
+### Proposal: `keepUpdated` element attribute
+The declarative version `updateRendering()` is an attribute called `keepUpdated`.
```
chrishtr, line 52:
```suggestion
The declarative version of `updateRendering()` is an attribute called `keepUpdated`.
```

explainers/update-rendering.md
```diff
@@ -16,9 +19,12 @@ words, if the website intends to show an element whose contents are currently
 skipped, then skipping work may cause jank when the contents are ultimately
 presented.
 
-### Proposal
-The proposal is to add the following function, which lets the user-agent know
-that the element’s contents should be updated in preparation for display:
+### Proposal: `updateRendering()` JavaScript function
```
chrishtr, line 22: WDYT of just removing updateRendering() for now, and just say "an imperative equivalent to keepUpdated could be added in the future".

chrishtr: I am pretty sure the attribute will be an easier sell for standards, just as it was for FB.

vmpstr: Sounds good.

explainers/update-rendering.md
```diff
@@ -1,13 +1,16 @@
-## `updateRendering()`
+## `updateRendering()` and `keepUpdated`
 
 ### TL;DR
 updateRendering() is an asynchronous JavaScript function on Element which
 requests that the rendering for the element be updated in preparation for
 display.
 
+The declarative counterpart is a `keepUpdated` attribute which lets the
```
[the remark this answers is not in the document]
vmpstr: Done.

vmpstr:
PTAL

explainers/update-rendering.md
```diff
@@ -16,38 +22,62 @@ words, if the website intends to show an element whose contents are currently
...
+  does _not_ include elements that themselves have `content-visibility: hidden`
+  style, since their rendering is updated and the attribute would cause the
+  subtree to be udpated as well.
 
+* Setting the attribute on an element that contains descendants with
+  `content-visibility: hidden`, `display: none`, or similar styles would not
+  cause such descendants to be updated. This is a consequence of the fact that
+  a fully updated parent element has all its rendering work completed without
+  updating such descendants. ([more
+  details](https://github.com/WICG/display-locking/issues/196))
```
chrishtr, line 82:
```suggestion
  details](https://github.com/WICG/display-locking/issues/196)). If such an update is desired, a `renderPriority` attribute should be set on such elements.
```

Should we add a sentence saying something like "a recursive `renderPriority` setting may be considered in the future?

explainers/update-rendering.md
```diff
@@ -1,13 +1,19 @@
-## `updateRendering()`
+## `renderPriority` attribute.
+
+_Note that issues referenced in this document can refer to either the
+`renderPriority` attribute or `updateRendering()` JavaScript API. The issues
```
chrishtr, line 4: is updateRendering mentioned anywhere in this doc now? If not, remove?

explainers/update-rendering.md
```diff
@@ -16,38 +22,62 @@ words, if the website intends to show an element whose contents are currently
...
+  for the purposes of User Agent prioritization.
+
+Note that the attribute values follow [the postTask API](https://wicg.github.io/scheduling-apis/#sec-task-priorities)
+example.
+
+
+### Interaction with `content-visibility`
+Since the User Agent typically keeps rendering state of subtrees up-to-date,
+this feature would be a no-op in a majority of cases. For instance, having
+`renderPriority` on a visible, on screen, element would not have to do any
```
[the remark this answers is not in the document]
vmpstr: I'm not sure about that.. I guess we haven't really discussed what userVisible means, but even if say it changes style and layout in separate frames, then that seems like it may be web breaking. I'm worried about making this enter the "render async" territory where things are visible but are not updated synchronously.

Or do you think that this is the opportunity to change this?

explainers/update-rendering.md
vmpstr: It's mentioned in the issues which I reference, so I just wanna keep this here as a clarification that we're talking about one and the same

vmpstr:
PTAL

explainers/update-rendering.md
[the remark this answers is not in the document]
chrishtr: > Or do you think that this is the opportunity to change this?

I think it's an opportunity to consider introducing async rendering across the board, not just in content-visibility. It would not be render breaking for existing content, because the `auto` behavior would be unchanged from today. We would also give UAs wiggle room to not implement it (e.g. we'd likely not do so in Chromium at first).

explainers/update-rendering.md
chrishtr: ok

explainers/update-rendering.md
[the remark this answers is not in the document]
vmpstr: I suggest that we move some of this discussion into an issue. I think that if a developer uses this in conjunction with `content-visibility`, and then at some point they remove `content-visibility`, they might not expect that although the element is now visible, it tends to be updated asynchronously:

```js
<div id=e style="content-visibility: hidden" renderPriority=background> ... </div>
...
e.style = "";
// e is now maybe asynchronous?
```

I think this text can change for sure, but coupling these two features in this explainer might complicated TAG/standards reviews. WDYT?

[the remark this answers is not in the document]
chrishtr: ok sure let's put it in an issue. File one and link to it, with a reference to the behavior being an open discussion point?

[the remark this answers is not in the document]
vmpstr: SGTM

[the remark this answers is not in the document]
vmpstr: Done.

chrishtr (approved):

PR #204, opened 2021-09-16
Update rendering example code

vmpstr:


vmpstr:
Please take a look.

explainers/update-rendering.md
```diff
@@ -68,6 +68,47 @@ rendering state is not updated to save CPU time. Adding `renderPriority` on
...
+Consider the following example.
+
+```html
+<div id=container style="content-visibility: hidden">
+  <!-- some complicated subtree here -->
+  ...
+</div>
+```
+
+Here the contents of the `#container` are not visible, and its rendering state
```
chrishtr, line 82:
```suggestion
Here the contents of `#container` are not visible, and its rendering state
```

explainers/update-rendering.md
```diff
@@ -68,6 +68,47 @@ rendering state is not updated to save CPU time. Adding `renderPriority` on
...
+is not necessarily up-to-date since the contents of the element are
+[skipped](https://www.w3.org/TR/css-contain-2/#skips-its-contents)
+
+This is typically done to avoid rendering work in this subtree. However, when
+the developer decides that the contents should now be visible, they remove
+`content-visibility: hidden` style. This causes all of the rendering to be
+updated in the subsequent frame. This work, in turn, can cause undue delay
+and visible jank on the page.
```
chrishtr, line 90:
```suggestion
(for example, some experimental data from Facebook indicates up to a [250ms](https://web.dev/content-visibility/#hiding-content-with-content-visibility:-hidden) delay due to this work in practice).
```

explainers/update-rendering.md
```diff
@@ -68,6 +68,47 @@ rendering state is not updated to save CPU time. Adding `renderPriority` on
...
+</div>
+```
+
+Here, the developer has added `renderPriority=background`, which means that they
+would like the User Agent to keep the rendering state of `#container`'s content
+to be kept up to date with a low priority.
+
+Assuming that the User Agent completes this work (i.e. it is not otherwise busy
+doing more important rendering work), then when the developer removes
+`content-visibility` style, the contents are displayed without delay or jank.
```
chrishtr, line 107:
```suggestion
the `content-visibility` style, the contents are displayed without delay or jank.
```

chrishtr: Add a caveat that this depends on the degree to which removing `content-visibility` causes a change of layout.

vmpstr:
PTAL

explainers/update-rendering.md
```diff
@@ -68,6 +68,52 @@ rendering state is not updated to save CPU time. Adding `renderPriority` on
...
+</div>
+```
+
+Here the contents of `#container` are not visible, and its rendering state
+is not necessarily up-to-date since the contents of the element are
+[skipped](https://www.w3.org/TR/css-contain-2/#skips-its-contents)
+
+This is typically done to avoid rendering work in this subtree. However, when
+the developer decides that the contents should now be visible, they remove
+`content-visibility: hidden` style. This causes all of the rendering to be
```
chrishtr, line 88:
```suggestion
the `content-visibility: hidden` style. This causes all of the rendering to be
```

chrishtr (approved):

PR #212, opened 2021-11-18
content-visibility:hidden-matchable -> hidden=until-found

josepharhar:


explainers/hidden-content-explainer.md
```diff
@@ -71,196 +75,152 @@ currently possible.
...
 
 Using this feature is a good optimization for low power mobile devices that
 don't want to render all of the content in the page but still making it
 accessible to features such as find-in-page. Making long articles with collapsed
 sections accessible to find-in-page will improve the user experience.
 
+## The `beforematch` event and other use cases
+
+The previous example doesn't need to use the proposed `beforematch` event
+because all other visual state in the example is controlled with HTML attribute
```
vmpstr, line 152: I suspect we'll have to update this.

There are sibling combinators, but they only match the second sibling, so I don't know if you can do this with css alone without some awkward positioning

explainers/hidden-content-explainer.md
```diff
@@ -71,196 +75,152 @@ currently possible.
...
+selectors in CSS. However, we should still notify the page that something is
+being revealed via the `beforematch` event. More complex use cases include
+carousels and tabs, where the revealing of one section means that the page must
+also hide another corresponding section, perhaps by applying the
+`hidden=until-found` attribute, in a `beforematch` event handler.
+
+## Comparison to the `<details>` element
+
+The [auto-expanding `<details>`
+feature](https://github.com/whatwg/html/pull/6466) works very similarly to this
```
vmpstr, line 162:
```suggestion
element](https://github.com/whatwg/html/pull/6466) works very similarly to this
```

explainers/hidden-content-explainer.md
```diff
@@ -71,196 +75,152 @@ currently possible.
...
-   [search-incremental.html](/resources/find-in-page/search-incremental.html),
-   firing beforematch on hidden content that stays hidden would be harder for
-   the user to notice.
-
-As for the ScrollToTextFragment case, the page can already guess with reasonable
-certainty based on the scroll offset which text is already highlighted. In
-addition, ScrollToTextFragment can only occur once at the beginning of a
-document's load, so we don't have to worry about the page building out a search
-term since there will only be one possible beforematch event.
+In order to mitigate find-in-page snooping from both `scroll` events and
```
vmpstr, line 182: I personally don't like the word "snooping" in this context, but it's up to you :)

explainers/hidden-content-explainer.md
```diff
@@ -71,196 +75,152 @@ currently possible.
...
+
+`hidden=until-found` will be implemented as a presentational style of
+`content-visibility:hidden` as opposed to the existing `hidden` attribute which
+applies a presentational style of `display:none`. This will improve the
+performance of searching the content inside the `hidden=until-found` subtree
+because searching for text in these areas requires computing style and layout.
+`content-visibility` saves this computed style and layout instead of
+`display:none` which deletes them, which means that once the first run of
+find-in-page or ScrollToTextFragment will compute these values, and subsequent
+runs can reuse the saved style and layout.
```
vmpstr, line 213: Would this be a good section to mention self-box is present and also c-v doesn't apply to non-atomic inlines

explainers/hidden-content-explainer.md
```diff
@@ -71,196 +75,152 @@ currently possible.
...
-Since changes to the element fragment trigger synchronous scrolling which we
-can't change without breaking the current behavior, the beforematch event must
-be fired either synchronously before the scroll or via a posted task on the
-event loop instead of the rAF->beforematch->rAF->scroll behavior we have for
-ScrollToTextFragment and find-in-page.
-This means that pages which need a full rAF in between the beforematch event and
-the scroll in order to reveal the content to scroll to may not be able to reveal
-the target content in time for the scroll.
+`hidden=until-found` subtrees will not be included in the accessibility tree
+because they internally use `contenet-visibility:hidden`.
```
vmpstr, line 218: I think we should be careful about stating that c-v is the reason for why they are not in the a11y tree. I think they aren't in the a11y tree because we don't want this content to be exposed to a11y since it goes against what the site author desires. The fact that we implement it using c-v hidden is tangential.

(also you have a typo in content-visibility)

explainers/hidden-content-explainer.md
```diff
@@ -71,196 +75,152 @@ currently possible.
...
+
+Websites are encouraged to provide a way to reveal `hidden=until-found` content
+without the use of find-in-page, ScrollToTextFragment, or element fragment
+navigation in order to ensure that all users can access the hidden content, like
```
vmpstr, line 222:
```suggestion
navigation in order to ensure that all users can access the hidden content using a method other than find-in-page, such as
```

explainers/hidden-content-explainer.md
```diff
@@ -271,15 +231,70 @@ only benefit I see to adding support to `window.find` would be to make it easier
...
 
+## Backwards compatability with the `hidden` attribute
+
+If a browser which doesn't implement this feature renders a `hidden=until-found`
+element, it will still apply the `display:none` style from the `hidden`
+attribute, which will still make the element hidden like the
+`content-visiblity:hidden` style that `hidden=until-found` is supposed to apply.
+This means that websites won't need to specially consider support for this
+feature between browsers, it just means that find-in-page and
+ScrollToTextFragment won't work as nicely in the browsers which don't support
```
vmpstr, line 242: This isn't true.

```
<style>
.hidden {
  width: 100px;
  height: 100px;
}
</style>
<div hidden=until-found class=hidden></div>
<!-- some other content -->
```

In this case, the box would take up 100x100 space since we're using content-visibility: hidden. We're hiding the content, not the element itself. With display: none, this would occupy 0 space

explainers/hidden-content-explainer.md
```diff
@@ -1,41 +1,47 @@
...
+  which is located inside an element with the `hidden=until-found` attribute.
+* There is a [scroll-to-text](https://github.com/WICG/ScrollToTextFragment)
+  navigation (`example.com/#:~:text=foo`), where the target text is located
+  inside a `hidden=until-found` element.
+* There is an element fragment navigation or change (`example.com/#foo`),
+  where the target is the element whose id matches the id in the fragment.
+* [`window.find`](https://developer.mozilla.org/en-US/docs/Web/API/Window/find)
+  found a text match located inside a `hidden=until-found`
+  element. We aren't certain if `window.find` will be supported for sure
+  yet, see the detailed section about it below.
```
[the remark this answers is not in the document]
josepharhar: Ok, I said that we won't support it until window.find get specced. How does that sound?

explainers/hidden-content-explainer.md
```diff
@@ -1,41 +1,47 @@
...
+
+If the matching text spans multiple `hidden=until-found` ancestors, only the
+first `hidden=until-found` will be removed and the `beforematch` event will only
+be fired on the first one.
+
+If there are multiple `hidden=until-found` elements in the flat tree ancestor
+chain of the target element to reveal, then all of them will be removed and all
+of them will receive the `beforematch` event.
+
+Note that the proposal for `hidden=until-found` and `beforematch` is still in
```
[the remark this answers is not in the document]
josepharhar: removed

explainers/hidden-content-explainer.md
josepharhar: Are you saying that we should add a new example that actually uses `beforematch`?

explainers/hidden-content-explainer.md
josepharhar: I'm not sure what you mean. Could you suggest an edit?

explainers/hidden-content-explainer.md
josepharhar: Ok, I merged it with some of the previous text. What do you think?

explainers/hidden-content-explainer.md
josepharhar: I see...
Should I just change the phrasing to suggest that `hidden=until-found` looks like `hidden` in most but not all cases...?

explainers/hidden-content-explainer.md
vmpstr: Sounds nefarious and informal. But it's fine, snooping it is!

explainers/hidden-content-explainer.md
vmpstr: I think it's a big enough distinction that it warrants a separate section. One of the biggest difference between something like `display: none` and `content-visibility: hidden` is its treatment of the element itself that is affected by the property. In `display: none` case, the element itself doesn't have a box so it isn't visible or styled or painted. In `content-visibility: hidden` case, the element itself is painted and drawn. You can check by adding a border on a `content-visibility: hidden` element -- the border is painted and drawn, even though the _contents_ of the element are not.

I feel like this distinction is important enough to point it out separately so that people reading this explainer are very well aware of it.

When it comes to updating this wording, then you can point to that section and say "`hidden` and `hidden=until-found` both hide the content. There, can, however be some visual differences. Link to the section"

explainers/hidden-content-explainer.md
vmpstr: See my other comment about this

explainers/hidden-content-explainer.md
vmpstr: That reads much better, thanks!

explainers/hidden-content-explainer.md
josepharhar: Now that I've actually tried it I see.
I updated it, and now it also uses beforematch.

I also removed the section which says that the example doesn't use beforematch and justifies why we need beforematch.

explainers/hidden-content-explainer.md
```diff
@@ -71,196 +72,155 @@ currently possible.
 .title::before {
   content: '⬇️ ';
 }
-.collapsed > .title::before {
+[hidden=until-found] > .title::before {
```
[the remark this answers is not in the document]
josepharhar: yeah it was very broken, i fixed it

explainers/hidden-content-explainer.md
josepharhar: I updated this section and added the `visibility` mitigation we discussed last week. How does it look now?

explainers/hidden-content-explainer.md
```diff
@@ -86,181 +87,142 @@ currently possible.
...
+searchable and automatically expanded.
 
 Using this feature is a good optimization for low power mobile devices that
 don't want to render all of the content in the page but still making it
 accessible to features such as find-in-page. Making long articles with collapsed
 sections accessible to find-in-page will improve the user experience.
 
+## Comparison to the `<details>` element
+
+The [auto-expanding `<details>`
```
vmpstr, line 160:
```suggestion
The [`<details>`
```

explainers/hidden-content-explainer.md
```diff
@@ -86,181 +87,142 @@ currently possible.
...
+element](https://github.com/whatwg/html/pull/6466) works very similarly to this
```
vmpstr, line 161:
```suggestion
element](https://html.spec.whatwg.org/#interaction-with-details) works very similarly to this
```

explainers/hidden-content-explainer.md
```diff
@@ -86,181 +87,142 @@ currently possible.
...
+feature. This feature empowers developers to build their own auto-expandable
+details elements and more by exposing the primitive features used to make the
+auto-expandable `<details>` element.
```
vmpstr, line 164:
```suggestion
`<details>` element be searchable.
```

explainers/hidden-content-explainer.md
```diff
@@ -86,181 +87,142 @@ currently possible.
...
+If the `beforematch` event handler
+[collapses](https://dom.spec.whatwg.org/#range-collapsed) the DOM range of the
+target text or changes the style to remove the target text from rendering, the
+browser will not scroll to the target text.
+
+## Usage of `content-visibility`
+
+`hidden=until-found` will be implemented as a presentational style of
+`content-visibility:hidden` as opposed to the existing `hidden` attribute which
+applies a presentational style of `display:none`. This will improve the
```
vmpstr, line 206:
```suggestion
applies a presentational style of `display:none`. This improves the
```

explainers/hidden-content-explainer.md
```diff
@@ -86,181 +87,142 @@ currently possible.
...
+performance of searching the content inside the `hidden=until-found` subtree
+because searching for text in these areas requires computing style and layout.
+`content-visibility` saves this computed style and layout instead of
+`display:none` which deletes them, which means that once the first run of
+find-in-page or ScrollToTextFragment will compute these values, and subsequent
```
vmpstr, line 211:
```suggestion
find-in-page or ScrollToTextFragment computes these values, the subsequent
```

explainers/hidden-content-explainer.md
```diff
@@ -86,181 +87,142 @@ currently possible.
...
+
+## Usage of `content-visibility`
+
+`hidden=until-found` will be implemented as a presentational style of
+`content-visibility:hidden` as opposed to the existing `hidden` attribute which
+applies a presentational style of `display:none`. This will improve the
+performance of searching the content inside the `hidden=until-found` subtree
+because searching for text in these areas requires computing style and layout.
+`content-visibility` saves this computed style and layout instead of
+`display:none` which deletes them, which means that once the first run of
```
vmpstr, line 210:
```suggestion
because `content-visibility: hidden`, unlike `display: none`, allows the rendering state
needed to perform a text search to be cached.
This means that once the first run of
```

explainers/hidden-content-explainer.md
```diff
@@ -86,181 +87,142 @@ currently possible.
...
+find-in-page or ScrollToTextFragment will compute these values, and subsequent
+runs can reuse the saved style and layout.
```
vmpstr, line 212:
```suggestion
runs can reuse the cached rendering state.
```

explainers/hidden-content-explainer.md
```diff
@@ -271,15 +233,83 @@ only benefit I see to adding support to `window.find` would be to make it easier
 to add WPT tests for beforematch, since we cant have WPTs for find-in-page.
 However, that would also require speccing `window.find`.
 
+## Backwards compatibility with the `hidden` attribute
+
+If a browser which doesn't implement this feature renders a `hidden=until-found`
```
vmpstr, line 238:
```suggestion
If a browser which doesn't implement this feature processes a `hidden=until-found`
```

explainers/hidden-content-explainer.md
```diff
@@ -271,15 +233,83 @@ only benefit I see to adding support to `window.find` would be to make it easier
...
+element, it will still apply the `display:none` style from the `hidden`
```
vmpstr, line 239:
```suggestion
attribute, it will still apply the `display:none` style from the `hidden`
```

explainers/hidden-content-explainer.md
```diff
@@ -271,15 +233,83 @@ only benefit I see to adding support to `window.find` would be to make it easier
...
+attribute, which will still make the element hidden like the
+`content-visiblity:hidden` style that `hidden=until-found` is supposed to apply.
```
vmpstr, line 241:
```suggestion
attribute, which will still make the element hidden.
```

explainers/hidden-content-explainer.md
```diff
@@ -271,15 +233,83 @@ only benefit I see to adding support to `window.find` would be to make it easier
...
+This means that hopefully websites won't need to specially consider support for
+this feature between browsers.
```
vmpstr, line 243:
```suggestion
This provides desired behavior on browsers that do not support `hidden=until-found`.
```

explainers/hidden-content-explainer.md
```diff
@@ -271,15 +233,83 @@ only benefit I see to adding support to `window.find` would be to make it easier
...
+
+However, `content-visibility:hidden` will still allow some parts of the
```
vmpstr, line 245: I'm not sure what "however" refers to here, this seems to be a separate topic from the previous paragraph

explainers/hidden-content-explainer.md
```diff
@@ -271,15 +233,83 @@ only benefit I see to adding support to `window.find` would be to make it easier
...
+`hidden=until-found` element to be rendered: the [generated
+box](https://drafts.csswg.org/css2/#visibility) of the element.
+
+As a mitigation for this, the following CSS should be included in the user agent
```
vmpstr, line 249: "To allow compatibility with hidden attribute, and to make adoption straight forward, the user-agent stylesheet must also include the following:"

I know this isn't a spec, but as a general comment in spec languages "should" typically means an optional thing which user-agents may omit (csswg tends to follow this: https://datatracker.ietf.org/doc/html/rfc2119, not sure about whatwg/html)

explainers/hidden-content-explainer.md
```diff
@@ -271,15 +233,83 @@ only benefit I see to adding support to `window.find` would be to make it easier
...
+stylesheet:
+```css
+[hidden=until-found] {
+  visibility: hidden;
+}
+[hidden=until-found] > * {
+  visibility: visible;
```
vmpstr, line 256: I forgot what if we said that things like
```html
<div style="visibility: hidden">
  <div hidden=until-found>text</div>
</div>
```
would be searchable with this style

explainers/hidden-content-explainer.md
josepharhar: I rewrote this section to be more clear, how does it look now?

explainers/hidden-content-explainer.md

explainers/hidden-content-explainer.md
josepharhar: ill discuss this some more in the design doc

josepharhar:
@vmpstr I updated the "Backwards compatibility" section as per our conversation

vmpstr (approved):

